PR #164, opened 2026-04-17
Multi-client integration test harness + DB resilience

actuallydan:
## Summary
- **Integration test harness** (`src-tauri/tests/flows.rs`): drives real `#[tauri::command]` functions through `tauri::test::get_ipc_response` with per-client `AppState` + `InMemoryKeystore` against a disposable Turso. 18 scenarios cover signup, group invites, join requests, MLS epoch ratchet across 9 users, remove-member access loss, randomized admin + membership churn, edit-across-membership-changes, envelope cleanup TTL/watermark, and the full DM lifecycle (multi-device round-trip, reject removes tree, re-invite rebuilds tree, block/unblock).
- **MLS ordering fix** (`83df6ef`): reconcile now stages commits remotely *before* merging locally, so a failed remote write can't leave the device one epoch ahead of the group.
- **Envelope cleanup semantics** (`e32978f`): the 30-day TTL and per-user watermark are now OR'd instead of AND'd, so idle watermarks can't pin old envelopes forever.
- **RemoteDb self-healing** (`e3f62d6`, `4c082c3`, `7998d1f`): `RemoteDb` holds its `Database` behind a `RwLock` so `reconnect()` can rebuild the handle after a libsql TCP reset / Hrana stream GC. `with_retry` wraps a single op in reconnect-and-retry-once; `ping` / `heal_if_stale` are on-demand primitives. Harness `wipe_remote` uses these to stabilize the ~4-minute serialized suite (previously flaked with "Connection reset by peer" mid-wipe). Deliberately no background polling anywhere in the app.
- **useNetworkStatus**: stripped of its 5s `setInterval` — Pollis has no centralized backend to ping, and the project rules out periodic polling.
- **Docs**: `.codesight/wiki/testing.md` covers harness architecture + every scenario; `.codesight/wiki/mls.md` documents the reconcile ordering invariant; `README.md` has the test-command table.

## Test plan
- [ ] `cargo test --features test-harness --test flows` — 18/18 pass (~195s locally)
- [ ] `cargo test` — unit tests still pass
- [ ] `pnpm test:e2e` — Playwright suite still green
- [ ] Manual: verify app still starts and login works end-to-end (no keepalive regression)


actuallydan:
## Follow-up commits on this branch

**26b5a02 — `fix(messages): make local DB the source of channel history`**

Architectural fix for disappearing messages. Local SQLite `message` table is now the authoritative source; remote `message_envelope` is a transient delivery queue gated by TTL + per-conversation watermark.

- New ingest helpers `ingest_channel_envelopes_inner` / `ingest_dm_envelopes_inner` pull envelopes → decrypt via MLS → `INSERT OR IGNORE` into local → advance watermark → run cleanup. Idempotent.
- `get_channel_messages` / `get_dm_messages` rewritten to ingest first, then page from local.
- Watermark advances past envelopes this device can't decrypt (wrong epoch). Rationale: an envelope encrypted to an epoch this device never had keys for is unreadable by this device — pinning GC on it would make it live forever for every future new member.
- Deleted the 4 remote-DB cursor SQL files (replaced by local SQL in `messages.rs`).
- Registered `ingest_channel_envelopes` / `ingest_dm_envelopes` as Tauri commands.
- Added `examples/migrate.rs` + `pnpm db:push` wiring (schema push to Turso).

All 18 harness tests pass (`cargo test --features test-harness --test flows`).

**8f327b4 — `fix(realtime): invalidate room cache for non-selected rooms`**

UX bug: UserB saw a preview update in the channels list when UserA sent a message, but entering the channel still showed stale data until Ctrl/Cmd+R.

Root cause: `useLiveKitRealtime`'s `new_message` / `edited_message` handlers only invalidated the channel/conversation query cache when that room was *currently selected*. For non-selected rooms the cache stayed fresh under React Query's 30s `staleTime`, so re-entering a channel served cached data with no refetch. The local DB was already up to date (via `useLastMessage`'s limit=1 ingest) — it was just React Query serving stale pages.

Fix: always invalidate the affected room's query. Mounted queries refetch immediately; unmounted ones just mark stale for the next mount. Unread/notification gating is unchanged.

Issue #184, opened 2026-04-23
User gets signed out inexplicably between sessions


actuallydan:
Sometimes the user is signed out and has to re-authenticate via email OTP. This happens:

- Sometimes after a crash
- Sometimes just periodically between sessions

The user is **not** signed out mid-session — but when opening the app the next day, re-OTP is sometimes required. Root cause unknown.

PR #197, opened 2026-04-25
fix(#194): make the PIN cryptographically load-bearing

### LINKED ISSUE #194: PIN: make cryptographically load-bearing (wire unwrap into app boot)
Follow-up to the PIN / #184 work on branch `fix/184-pin-and-session-cleanup` (6 commits, merged separately). Today the PIN is a UX gate + a storage scheme: wrap/unwrap works, the screens exist, `AppState.unlock` populates on `set_pin` / `unlock`. But the rest of the app still reads the **unwrapped** `db_key_{uid}` / `account_id_key_{uid}` keystore slots, so entering the wrong PIN rejects the frontend without actually gating decryption. This issue is to finish the job.

## Goal

After this lands:
- `db_key` and `account_id_key` live **only** as ciphertext in `db_key_wrapped_{uid}` / `account_id_key_wrapped_{uid}`.
- The app cannot open the local SQLCipher DB or sign anything until `AppState.unlock` is populated by a correct PIN entry.
- Wrong PIN means no data is decrypted, period.

## Where to start

Read these first:
- `PIN_AND_SESSION_CLEANUP.md` at repo root (design doc, gitignored) — section \"What gets deleted\" enumerates the exact grep targets.
- `src-tauri/src/commands/pin.rs` — wrap/unwrap + `UnlockState`. `AppState.unlock: Arc<Mutex<Option<UnlockState>>>` is already populated by `set_pin` and `unlock`.
- Commits on `fix/184-pin-and-session-cleanup`: `8eae93a`, `f46be90`, `2a4212c`, `3578122`, `1edbdbf`, `ed00bef`.

### Backend
- [ ] `state.rs::load_user_db` — stop reading the `db_key_{uid}` keystore slot. Take the already-unwrapped key from `AppState.unlock` (return `Err` if locked), or accept it as a parameter. Keep the mls_kv empty-check + welcome-reset logic.
- [ ] `state.rs::AppState::load_user_db` callers — trace every call site (`auth.rs`, `device_enrollment.rs`, others) and make sure they only run in a post-unlock context.
- [ ] `commands/auth.rs::verify_otp` — stop calling `load_user_db`. Keep: `generate_account_identity` (still generates the key, but returns raw bytes), `upsert_account`, returning profile with `new_secret_key`. `register_device` can probably still run here — it only touches the remote DB, not the local one. Everything that needs `db_key` / `account_id_key` moves to after `set_pin`.
- [ ] `commands/auth.rs::dev_login_inner` — mirror the verify_otp changes.
- [ ] `commands/auth.rs::get_session` (resume) — same: don't open the local DB here. Route should be `get_unlock_state` → pin-entry → `unlock` → open DB.
- [ ] `commands/account_identity.rs` — rewrite the 6 keystore sites:
  - `generate_account_identity` (line 213): stop storing the raw key in `account_id_key_{uid}`; return the bytes to the caller, which hands them to `set_pin` for wrapping.
  - `reset_identity_and_recover` key write (line 372): same treatment.
  - `load_account_id_key` (line 242/393): read from `AppState.unlock.account_id_key`; error cleanly if locked.
  - `has_local_account_identity` / `has_matching_local_account_identity` (line 258): these are metadata checks. Either look at `account_id_key_wrapped_{uid}` presence, or thread `AppState` through so they can check `unlock`.
  - `wipe_local_account_identity` (line 277): delete `account_id_key_wrapped_{uid}` (and clear `unlock` for that user).
- [ ] `commands/device_enrollment.rs` — 2 raw `account_id_key` store sites (lines 337, 620). New-device enrollment currently writes the raw unwrapped key. Needs the same rewrap-via-set_pin pattern.
- [ ] `commands/pin.rs::set_pin` (initial-set path) — after wrap succeeds, **delete** the legacy unwrapped slots (`db_key_{uid}`, `account_id_key_{uid}`). Currently it leaves them alone (stage 3 safety measure, no longer needed once callers are rewired).
- [ ] `commands/pin.rs::unlock` — after populating `AppState.unlock`, call `state.load_user_db_with_key(user_id, &db_key)` (new helper) so the DB opens as part of the unlock.
- [ ] `commands/pin.rs::lock` — already drops `AppState.unlock`. Add `state.unload_user_db().await` so the DB handle closes too.

### Frontend (`frontend/src/App.tsx`)
- [ ] Signup flow (`handleAuthSuccess` branch 1 → `handleSecretKeySaved`): verify_otp returns, secret key shown, then `set_pin`. After set_pin the DB is open — THEN call `initialize_identity` (MLS key package). Today `initialize_identity` is in `completeSignIn`. Keep it there, but ensure `completeSignIn` only runs after PIN unlock.
- [ ] Resume flow (`checkStoredSession`): `getUnlockState` first. If `pin_set && !is_unlocked` → pin-entry. Only after `unlock` (which now also opens the DB) call `completeSignIn`.
- [ ] Returning-user OTP flow (`handleAuthSuccess` branch 3): already routes to pin-entry or pin-create. Confirm it still works end-to-end after backend changes.

### Tests (`src-tauri/tests/flows.rs`)
- [ ] `TestClient::sign_up` — after `verify_otp`, call `set_pin` before `initialize_identity`. Pick a fixed test PIN like \"0000\". Without this, every single test breaks because `initialize_identity` needs the DB which needs a PIN.
- [ ] Add a test: sign_up → lock → (assert DB ops fail) → unlock with correct PIN → (assert DB ops succeed).
- [ ] Add a test: sign_up → lock → unlock with WRONG PIN → DB ops still fail → verify ciphertext is unreadable.

### E2E / mock
- [ ] `frontend/src/__mocks__/tauri-core.ts` already fakes `get_unlock_state` as `{ pin_set: true, is_unlocked: true }`. That keeps e2e tests passing without PIN entry. Confirm still correct after real flow is wired.

## Gotchas

- **`register_device` vs. DB open.** register_device only writes to remote Turso, no local DB needed. It can run before PIN. The account_id_key signing for device certs — double-check that signing path; if it signs with the account_id_key, it needs unlock first.
- **`ensure_mls_key_package`** reads account_id_key + opens mls_kv in local DB. Must run post-unlock.
- **Migration for existing users.** Someone who upgrades to a \"PIN load-bearing\" build while still having the legacy unwrapped `db_key` in their keystore: their first `set_pin` wraps what's there, then deletes the legacy slot. Same path as today's stage-4 flow, just now the deletion actually happens.
- **Forgotten PIN recovery.** `handlePinForgot` in App.tsx routes to the enrollment gate (Secret Key recovery). That path has to work: Secret Key decrypts the server-side wrapped account key, user lands at pin-create, `set_pin` wraps under the new PIN. Test this end-to-end.
- **`AppState.unlock` held behind a Mutex.** Every command that currently does `state.keystore.load_for_user(\"account_id_key\", ...)` becomes `state.unlock.lock().await.as_ref().ok_or(\"locked\")?.account_id_key.clone()`. Consider adding a helper like `state.require_unlock(user_id)` that returns `UnlockState` or a typed \"locked\" error.

## Definition of done

Grep targets that should return zero hits after the change:
- `load_for_user.*\"db_key\"` (only the legacy/migration cleanup path in `set_pin` allowed)
- `store_for_user.*\"account_id_key\"`
- `load_for_user.*\"account_id_key\"` (only the legacy fallback in `set_pin` allowed)

Flows suite still green (expect ~15-20 min runtime; see also #193 for cutting that).

actuallydan:
## Summary

Closes #194. Builds on #195 (the PIN UX gate + storage scheme). After this lands:

- `db_key` and `account_id_key` exist on disk **only as ciphertext** in the `*_wrapped_{uid}` keystore slots.
- The local SQLCipher DB cannot open until `AppState.unlock` is populated by a correct PIN entry, by `set_pin` on a fresh signup, or by an enrollment unwrap that hands raw bytes directly to `unlock` without ever writing them to the keystore.
- Wrong PIN → no plaintext on disk or in memory. `lock` closes the DB handle.

The pivot point: `AppState.unlock` becomes the single in-memory source of truth for both keys, regardless of where the bytes came from. Every DB open in the app now flows through `state.load_user_db_with_key`, with the key sourced from there. The legacy keystore-reading `state.load_user_db` is gone.

## Commit-by-commit

1. `refactor(#194): add load_user_db_with_key` — new entry point that takes the unwrapped `db_key` as a parameter; legacy entry point unchanged for now.
2. `refactor(#194): account_identity helpers read from AppState.unlock` — `load_account_id_key`, `has_local_account_identity`, `has_matching_local_account_identity`, `wipe_local_account_identity`, `sign_device_cert` all take `&AppState` and prefer `unlock` over the keystore.
3. `refactor(#194): generate_account_identity / reset_identity write to unlock` — fresh signup and identity reset stop writing raw bytes to the keystore. New `unlock_state_with_fresh_db_key` helper produces an `UnlockState` with a freshly generated `db_key` alongside the account_id_key.
4. `feat(#194): set_pin / unlock / lock are now load-bearing` — `set_pin` sources from `unlock` first (legacy keystore fallback for upgraders), wraps, deletes legacy slots, opens the DB, publishes the device cert. `unlock` does the same plus migrates any #195-vintage residue away. `lock` closes the DB.
5. `feat(#194): auth.rs stops opening the local DB pre-PIN` — `verify_otp` / `dev_login_inner` / `get_session` no longer call `load_user_db`. `register_device`'s tail call to `ensure_device_cert` moves into `set_pin` / `unlock`. Legacy `state.load_user_db` deleted.
6. `feat(#194): enrollment paths populate AppState.unlock; finalize is deferred` — `poll_enrollment_status`, `recover_with_secret_key`, `reset_identity_and_recover` route through `AppState.unlock` instead of the keystore. New Tauri command `finalize_device_enrollment` runs the cert / key-package / external-join work after `set_pin` opens the DB.
7. `feat(#194): App.tsx finalizes enrollment after pin-create` — `handlePinCreated` calls `api.finalizeDeviceEnrollment` between PIN creation and `completeSignIn`. Idempotent for fresh signup, load-bearing for enrollment paths.
8. `test(#194): flow harness sets PIN; load-bearing property tests` — `TestClient::sign_up` calls `set_pin(TEST_PIN)` between `verify_otp` and `initialize_identity`. `enroll_second_device` mirrors the App.tsx pattern. Two new property tests: `pin_locks_db_access` (lock genuinely closes the DB) and `wrong_pin_keeps_db_locked` (wrong PIN doesn't pry the file open and doesn't corrupt wrapped material).
9. `docs(#194): wiki updates for the load-bearing PIN cutover` — pin-design.md gets a Status section noting #184/#195/#194 progression. commands.md gets a `pin` section and the new `finalize_device_enrollment` command.

## Migration

Two upgrade paths covered:

- **Pre-PIN install** (no `pin_meta`, plaintext `db_key_{uid}` + `account_id_key_{uid}` slots): boot routes to `pin-create`. `set_pin` initial-set falls back to the legacy keystore slots, wraps them, **deletes** the slots after the wrap succeeds.
- **#195-vintage install** (PIN was set, but unwrapped slots also still exist alongside the wrapped ones): first successful `unlock` under the new build deletes the legacy slots.

## Definition of done

Grep audit, all clean (legacy reads only inside `pin.rs`'s migration fallback or the test module):

- `load_for_user.*"db_key"` → only `pin.rs`'s legacy-fallback path (via `DB_KEY_SLOT_LEGACY` constant).
- `store_for_user.*"account_id_key"` → zero hits in production code.
- `load_for_user.*"account_id_key"` → only `pin.rs`'s legacy-fallback path.

## Test plan

- [x] `cargo check --tests --features test-harness` clean.
- [x] `pnpm exec tsc --noEmit` clean.
- [x] Full `cargo test --features test-harness --test flows` suite green (22 tests, including 2 new load-bearing property tests, in 347s).
- [ ] **Manual smoke (on the reviewer):**
  - Fresh signup → SaveSecretKeyScreen → PinCreate → main app → send a message → restart → PinEntry → still see the message.
  - Wrong PIN flow shows "incorrect; N attempts remaining"; correct PIN unlocks.
  - Lock from settings (or via dev menu) → chat is gated → unlock → chat resumes.
  - Second-device enrollment → approval from device A → device B routes to pin-create → set PIN → main app loads → send a DM → device A receives.
  - Upgrade from a #195-vintage build (legacy slots still present): sign in, enter existing PIN → `unlock` migrates legacy slots away → restart → `unlock` finds only wrapped slots.
- [ ] `pnpm test:e2e` (Playwright suite) — mock stubs `finalize_device_enrollment` so the gate-skip path remains intact.

Marking as draft until manual smoke is signed off.


actuallydan:
Thanks Jeff :+1:

Issue #918, opened 2026-08-16
The flows harness hand-mirrors the DS router and is missing ~10 routes, so whole features have no integration coverage


actuallydan:
Found during #875 phase 2b.

`src-tauri/tests/flows/harness.rs` re-declares the DS's routes by hand: 61 against `pollis-delivery/src/lib.rs`'s ~73. Anything absent 404s inside the integration suite, so a feature can look covered while its flows tests cannot reach it at all.

`/v1/invite-links/{create,revoke}` were missing entirely — #847 invite links 404'd for every flows test. Those two were added in #875 because a test there needed them. **Still missing:**

- `/v1/emoji/{create,remove,gc}` — #848 custom emoji has **zero** flows coverage
- `/v1/invite-links/redeem`
- `/v1/welcomes/resubmit`
- `/v1/livekit/*`
- `/v1/r2/presign`
- `/v1/turso/token`
- `/v1/retention/metrics`
- `/v1/config`

Two things to fix, and the second matters more:
1. Add the missing routes and the flows coverage they unblock (emoji especially — a whole shipped feature with no integration test).
2. Stop the list being hand-maintained. This is the same duplication shape as the typed-DS-client item in #875: two hand-mirrored lists that drift silently, where the failure mode is a 404 that reads as a test gap rather than a mismatch. Mount the real router, or generate both sides from one declaration.

actuallydan:
Resolved by #920 (#875's typed-DS-client phase), not partly — fully.

The harness's only reason to mirror the router was the DB handle type. `RemoteDb::shared_database()` + `Db::from_shared()` let both sides wrap the same libsql `Database`, so `src-tauri/tests/flows/harness.rs` now mounts `pollis_delivery::build_router_with_state` verbatim — **1,651 lines of hand-mirrored routing deleted**. `/v1/commits` remains overridden on an outer router with `fallback_service` → the real one, so fault injection survives.

Every future endpoint reaches production code with no harness edit, and `pollis_api::ENDPOINTS` is generated by the same macro that writes the `DsRequest` impls, so the table cannot drift from the types. Two coverage tests assert every declared endpoint is routed and reachable; both were verified to fail when routes were removed.

The missing routes are in, including `/v1/emoji/*` — #848 custom emoji had zero integration coverage and now has flows tests.

Issue #919, opened 2026-08-16
foreign_keys=OFF reaches only the first DS connection, so test results depend on checkout nesting


actuallydan:
Found during #875 phase 2b.

`pollis-delivery/src/db.rs`'s `Db::connect_local` sets `foreign_keys=OFF` on the connection it primes and parks. `Db::conn()` is sync, so a **nested** checkout builds a fresh connection where libsql's local backend defaults foreign keys **ON**. Probed directly: `conn a: foreign_keys=0`, `conn b: foreign_keys=1`.

Production runs against remote Turso, where FKs are off — so this is test-only. But it means DS fixtures pass or fail depending on whether a checkout happens to be nested, which is the kind of nondeterminism that gets "fixed" by loosening a test.

#875 worked around it (explicit pragma where a raw builder is used, referentially-complete seeds elsewhere) and documented both patterns in the wiki. The real fix is an async `conn()` so every checkout runs the same pragmas — that is #777 connection-pool territory.

actuallydan:
Fixed in #955 (`9c6909e9`). `Db::conn()` is now async (317 call sites) so it stamps every connection it builds, matching the shape `RemoteDb::conn` already had, and two now-redundant per-test workarounds were retired. Worth recording: chasing this is what surfaced that production Turso runs with foreign keys OFF, which led to #959 — deletion had never actually cascaded.

Issue #917, opened 2026-08-16
Several group reads take no caller at all, exposing any group's roster, channels and emoji to anyone


actuallydan:
Found during #875 phase 2b while unifying the group-admin authz preflight, and deliberately left out of that change (it fixes *inconsistent* authz; this is *absent* authz, and it needs its own review).

These take no caller identity and perform no membership check:

- `get_group_members` — returns **any group's full roster** to anyone who names the group id
- `list_group_channels` — returns any group's channel list
- `list_group_emoji` — returns any group's custom emoji
- `search_group_by_slug` — enumerates the whole `groups` table
- DS `apply_create_join_request` — no membership check at all

Every client holds a whole-DB read-only Turso token, so group ids are enumerable; this is reachable, not theoretical. The stated trust model treats the network, Turso, the DS and operators as untrusted but does not license one *user* to read another group's membership. A roster is exactly the social-graph metadata the product exists to protect.

Note `list_group_emoji` is more nuanced: #848 deliberately allows rendering an emoji from a group you are not in (the content hash resolves without a membership check, by design). *Listing* a group's emoji set is a different act from resolving one hash, and the design note for #848 should be read before deciding the rule.

Wants a deliberate read-authz policy per endpoint rather than a blanket guard — decide for each what a non-member should get (403 vs empty vs a redacted shape), then enforce it at the chokepoint the way #875 did for the admin preflight.

Issue #916, opened 2026-08-16
Batched IN (…) queries have no chunking and will hit SQLite's parameter limit


actuallydan:
Found during #875 phase 3.

The N+1 collapses added in #875 build `IN (…)` clauses from a caller-supplied list with no chunking. SQLite's default `SQLITE_MAX_VARIABLE_NUMBER` is 999 in older builds (32766 in newer), so a query over ~1000 channels or roster members would fail outright rather than degrade.

Not reachable in practice today — no group has anywhere near that many channels or members — and it matches the existing norm in `batch_check_and_pin_account_keys`, which has the same shape. Filed so the ceiling is written down rather than discovered later by whoever first builds a group big enough.

Fix: chunk at a conservative size (e.g. 500) in one shared helper, and use it for every batched read including the pre-existing one.

PR #920, opened 2026-08-16
fix(backend): close the #875 audit — drifted duplication, hot paths, and the schema fixtures

### LINKED ISSUE #875: Backend (Rust) performance and duplication audit — ~58 findings, several live bugs
Audit of all Rust in the monorepo for network round trips, crypto, media, hot paths and cross-crate duplication. **~58 distinct findings.**

The headline: **several of these are live bugs, not slow paths.** Duplication in this codebase has already drifted into behavioural divergence — the same logic exists in two places and the copies no longer agree. Those are listed first because they cost correctness, not milliseconds.

## Correctness bugs caused by drifted duplication — fix first

- [ ] **A transparency signing-key rotation makes the shipped client cry wolf.** `PublicKeyDoc::active_keys` (`verifiable-log-serve/src/bundle.rs:191`) applies the rotation-overlap key set and has exactly **one** caller. The other three verify paths — `account.rs:148-151`, `group.rs:170-172`, `release.rs:158-160` — rebuild a `Bundle` with `retired_keys: Vec::new()`, each carrying an identical comment claiming the overlap set "is applied when selecting the verifying key", which on those paths it is not. During an overlap `pollis-verify verify` passes while `account`/`group`/`release` return `chain_valid = false`, so `self_audit_account_key`, `audit_peer_account_key` and `verify_own_build` raise a hard `AuditStatus::Alarm` **against an honest log**. Route every path through `active_keys` + `verify_any_with_context`.
- [ ] **`/v1/messages/delete` hard-403s whenever DS auth is disabled.** The server needs `actor_id` as the no-auth fallback (`pollis-delivery/src/messages.rs:697-698` → `writes.rs:196-199` returns `Forbidden` when both are `None`); both client call sites omit it (`edit_delete.rs:145-149`, `:227-231`) and `grep '"actor_id"' pollis-core/src` returns **zero** hits, while every sibling endpoint sends it.
- [ ] **RFC3339 precision was hardened on one side only.** `pollis-delivery/src/messages.rs:400` deliberately uses `SecondsFormat::Nanos`, with a comment at `:386-398` explaining that a whole-second timestamp sorts below a fractional one (`'+'` < `'.'`) and silently buries admin delete tombstones under recipients' watermarks. pollis-core writes the same lexically-compared columns with plain `to_rfc3339()` (`AutoSi`, which emits no fraction at exactly zero nanos) at **21 sites**.
- [ ] **pollis-core declares two private `PublicKeyDoc` stand-ins** (`commands/transparency.rs:553-555`, `:580-582`) that keep only `public_key` and drop the `keys` array — which is what makes the rotation bug above unfixable client-side. It already depends on `verifiable-log-serve`, which re-exports the real type. The doc comments there still say "Ed25519"; the key is ML-DSA-44.
- [ ] **Only one of the two signed-envelope verifiers has the cross-artifact guard.** `net/directory.rs:219-256` and `pollis-relay/src/policy.rs:360-403` are line-for-line identical against the same pinned key, except the relay copy checks `kind`. `Directory` has no `kind` field at all; artifact swapping is blocked today only by an incidentally missing `#[serde(default)]`.

## Network round trips

- [ ] **Build the HTTP client once.** The DS constructs a fresh `reqwest::Client` on **every** outbound call, and pollis-core does the same at `net/directory.rs:193` and `commands/voice/lifecycle.rs:149` — the exact anti-pattern `pollis-relay/src/http.rs:9-16` documents as costing **3–4.5 s per request**.
- [ ] **Stop reloading the whole MLS group per envelope.** `try_mls_decrypt` performs ~12 SQL reads plus a ratchet-tree JSON parse for *every* message decrypted.
- [ ] **Collapse the N+1 reads** — per-member/device/message SELECTs that should be one `IN (…)`, and sequential `await`s that should be `join_all`.
- [ ] **Cache stable data** — profiles, device certs and key packages are re-fetched rather than held.
- [ ] **Adopt one retry/backoff policy** — there are five different schedules with no shared definition.

## Crypto and hot paths

- [ ] **Cache device-cert verification verdicts** instead of re-verifying the same cert repeatedly, and hoist signature verification out of loops.
- [ ] **Audit lock-across-`await` sites** — a correctness smell as much as a performance one.
- [ ] **Stop cloning large buffers** (`Vec<u8>` ciphertext, message bodies) where `&`/`Arc` would do.

## Duplication that should be shared

- [ ] **Add a `pollis-schema` crate.** pollis-delivery hand-rolls **17 test schemas containing 11 different `user_device` definitions**, because it deliberately does not depend on pollis-core and so cannot use the public `BASELINE_SQL`. This has already drifted: `tests/retention.rs:27-28` gives `group_member` no `role` and no PK; `tests/key_packages.rs:52-57` uses a composite PK where the baseline uses `device_id` alone; `src/messages.rs:1833` and `:2641` disagree **within one file**. Missing PKs let those tests create duplicate membership rows production forbids — directly weakening the retention tests. `pollis-device-cert` is the precedent.
- [ ] **Type the DS client.** 64 endpoint pairs are mirrored as typed server structs against untyped `json!{}` client literals — the shape that produced the `actor_id` bug above.
- [ ] **Extract the group-admin authz preflight** — the same 8 lines are copied **12 times** across `commands/groups/*` and `messages/edit_delete.rs:121`, with four different "not a member" strings for one condition, including one that silently returns `Ok(Vec::new())`. pollis-delivery already has the `group_role`/`is_admin` pair.
- [ ] **Give the monitor CLI a context flag.** `verifiable-log/src/bin/monitor.rs:144` hardwires the commit-log domain, so it cannot verify account-keys or binaries bundles despite the builder README documenting that flow — two builder tests assert `verify` must return *false* and re-implement the loop. Its private `Bundle` also lacks `retired_keys`, so serde drops the field and the monitor fails every STH minted under a retiring key.
- [ ] **Merge the duplicated account-creation-from-email** — `auth.rs:918-941` ≡ `otp.rs:414-445`, already minor-drifted.
- [ ] **Consolidate `now_unix` / `hex_lower`** and the two 1000–2500-line test harnesses, where `build_router_with_state` and `Config::for_test()` already exist.
- [ ] **Delete two stale comments** — `pollis-delivery/Cargo.toml:37` cites a `livekit_jwt` module that no longer exists; `pollis-delivery/src/groups.rs:93-96` claims to mirror a deleted pollis-core function.

## Deliberate duplication — do NOT merge
The `is_member` mirror across the trust boundary (worth a parity test, as `anchor.rs:74-77` already does), the DS canonical signing message, and `pollis-device-cert` / `pollis-capture-proto` / `verifiable-log::merkle`, which are genuinely single-homed. `src-tauri` shims are clean pure forwarders.

Findings are labelled measured vs reasoned in the audit; the only hard number quoted here (3–4.5 s per client construction) is from the codebase's own documentation.

actuallydan:
Closes #875.

Five phases, 26 commits. The audit's thesis held up: **duplication here had already drifted into behavioural divergence**, and most of the value was correctness, not milliseconds.

### Correctness bugs fixed

- **A key rotation made the client accuse an honest log of tampering.** `active_keys` had one caller; `account`/`group`/`release` each rebuilt a bundle with `retired_keys: Vec::new()` under a comment claiming otherwise. Measured before the fix: `pollis-verify` **ok**, the other three **"STH signature is invalid"** on the same honest mid-rotation log. All four now share one candidate-derivation path.
- **The monitor CLI told auditors every non-commit-log tree was lying** — account-keys, binaries, and any head under a rotating key all exited 1, against the exact fixtures its README documents. Its private `Bundle` dropped `retired_keys` at the serde layer.
- **One email could create two accounts.** The DS trimmed before touching `users`; the client copy did not. `users.email` is `NOT NULL UNIQUE`, so `" a@x.com "` missed the row `"a@x.com"` owns and inserted a second account — `UNIQUE` cannot catch it, the strings genuinely differ. Fixed at the function holding the INSERT, because the desktop and TUI harnesses bypass the handler.
- **Four endpoints hard-403'd with DS auth disabled**, not the two reported: both delete branches plus emoji create/remove.
- **A mutex held across three awaits.** `if let Some(x) = m.lock().await.clone()` holds the guard to the end of the body — `ds_post` takes that same non-reentrant mutex, so this was one refactor from a self-deadlock with no compiler diagnostic. A source-shape guard now forbids the pattern.

### Performance

MLS decrypt was **12 SQL reads per message**; it is now 12 per *epoch* — 200 messages went 2,400 → 12 reads. Safe because decryption mutates only the message-secrets store, which openmls writes through before returning; a decryptor never spans a commit. Plus four N+1 collapses on the receive path, one shared HTTP client (four live construction sites), and one jittered backoff definition — two refresh loops were deriving their sleep from a signed artifact's `expires_at`, i.e. **the same instant for every client**, so the fleet was in permanent lockstep.

**No new caches of identity or crypto material**, deliberately: a memoized `account_id_pub` would verify a cert forged under a key that was rotated *because* it was compromised, and key packages are single-use by protocol.

### Schema fixtures

Not 17 hand-rolled schemas — **22 blocks, 60 structurally distinct definitions of 27 tables**, plus 5 more in pollis-core the audit missed. 12 had hard drift. Two tests were provably weaker than they claimed: one passed on `"no such column"` rather than the poison `CHECK` it was written for, and another seeded with `INSERT OR IGNORE` that inserted **nothing** against real constraints, making 7 reference-counting tests vacuous. New `pollis-schema` crate owns the migrations; the DS takes it as a **dev-dependency only**, so the binary still carries no schema and still does not depend on pollis-core.

### Authz

14 copies of the admin preflight (not 12) collapsed to one module, with one wording per condition instead of four. **Behaviour change:** two commands that returned `Ok(Vec::new())` to non-admins now error — an empty list is byte-identical to "nothing here", so a UI cannot tell a permission boundary from an empty page. The frontend already gates both on role, so nothing user-visible changes.

### Findings overturned

- **The RFC3339 "message loss" finding was a false positive.** `to_rfc3339()` omits the fraction only when nanos are genuinely zero — it never truncates — so a fraction-less stamp denotes `.000000000` and sorting it below fractional stamps in the same second is *correct*. 5 sites that write `sent_at` were routed through one chokepoint anyway; the other 20 were left, with reasons.
- **"Five retry schedules" was wrong** — there are eight, deliberately different. The real finding is that `RemoteDb::with_retry` has **zero production callers** while five comments cite it (#914).

### Verification

pollis-core **590** · pollis-delivery **271** · pollis-relay **106** · verifiable-log **56** / serve **64** / builder **64** · pollis-schema **3** · `flows` **86 passed, 0 failed** (752s). All 0 failed. Fix-proving tests were each verified to fail with their fix reverted; regression guards are labelled as guards, not proofs.

### Filed, not silently dropped

#908 watermark seed format · #910 dev-only remote INSERT · #912 no clippy on the pinned toolchain · #913 DS has no outbound timeouts · #914 unused retry helper · #915 3× attachment copies in memory · #916 unchunked `IN (…)` · #917 **group reads with no caller identity at all** · #918 flows harness missing ~10 routes (#848 emoji has zero coverage) · #919 per-connection FK pragma.

Not done, deliberately: the typed DS client (64 endpoint pairs) — it is the same shape as #918 and wants one declaration feeding both sides rather than a hand-merge.


actuallydan:
**Not ready to merge.** The typed DS client — the last outstanding item in #875 — is still in flight on `fix/875-typed-ds-client` and will be integrated into this branch before merge.

Opening the PR without it would have left part of its own ticket undone, which CLAUDE.md rules out and is how half-landed work gets forgotten. It also plausibly closes #918 outright: if the flows harness can mount the real router instead of re-declaring 61 routes by hand, the drift that hid #848's missing coverage stops being possible.

actuallydan:
**Ready to merge.** The typed DS client is integrated — the last outstanding item in #875.

It found **16 more endpoints** with the same actor bug (20 total), including all four broker endpoints, so voice, screenshare and media were broken for any self-hosted deployment running without DS auth. A missing field is now a compile error rather than a runtime 403.

**#918 is fully resolved, not patched:** the flows harness mounts the real router instead of hand-declaring it — 1,651 lines of mirrored routing deleted, and every future endpoint reaches production code with no harness edit.

Final: pollis-core **593** · pollis-delivery **274** · pollis-relay **106** · pollis-api **6** · pollis-schema **3** · verifiable-log **56** / serve **64** / builder **64** · `flows` **91 passed, 0 failed** (732s). Zero failures anywhere.

New tickets from this phase: #921 (`POLLIS_DS_REQUIRE_AUTH` defaults OFF — now more dangerous, since the broken endpoints were an accidental safety net), #922 (response bodies are the same bug, other direction), #923, #924, #925.

PR #949, opened 2026-08-17
Encrypt the desktop file keystore at rest and choose the backend at runtime

### LINKED ISSUE #882: Encrypt the desktop file keystore at rest so headless installs are not forced to choose
Follow-up to #879 and #881.

**The trade as it stands.** `pollis-core`'s keystore backend is selected at compile time (`keystore.rs:366` vs `:561`): with `os-keystore` the OS keychain, without it a **plaintext JSON file** whose own comment says "debug builds only". #879 turned the feature on for `pollis-tui` because the shipped CLI was writing keys in the clear — removing one of the four components the threat model calls trusted, and leaving a 4-digit PIN as the only protection once the file is copied.

That was the right call for confidentiality and the wrong one for the CLI's likely home: on a box with no secret-service — a server over SSH — `keyring` fails, so the terminal client cannot start. #881 fixes the *build*; nothing fixes that.

**The fix.** Encrypt the file backend at rest instead of leaving it plaintext. The machinery already exists and is proven: on Android and iOS that same file already holds AES-GCM ciphertext via `super::android_kek` / `super::ios_kek` (`keystore.rs:416-423`). Desktop is the outlier.

With that in place the backend choice stops being a security cliff:
- OS keychain where one exists (best — the OS guards the key)
- encrypted file where one does not (acceptable — no plaintext on disk)
- plaintext never

**Design question to settle first:** where the desktop KEK comes from. Mobile uses a platform keystore, which is the thing we are assuming absent. Options are a PIN-derived KEK (ties key confidentiality to the PIN's entropy, which is weak for 4 digits) or a machine-bound secret. Decide and record the reasoning before implementing — this is the part that determines whether the encrypted file is meaningfully better than plaintext or only cosmetically so.

Ship with a test that the release desktop path never writes plaintext, so the regression #879 fixed cannot recur silently.

actuallydan:
Closes #882.

The desktop file keystore is now AES-256-GCM at rest, and the backend is chosen
at **runtime** instead of at compile time. Same shipped binary: OS keychain on a
desktop, machine-bound encrypted file on a box with no secret-service, plaintext
nowhere. That dissolves the trade #879 had to make.

## The KEK, and what it actually buys

`HKDF-SHA256(ikm = platform machine ID, salt = fresh 16 bytes per write)`, where
the machine ID is `/etc/machine-id` → `/var/lib/dbus/machine-id` (Linux),
IOPlatformUUID (macOS), MachineGuid (Windows). Envelope is the same
`kek_envelope` iOS already uses, so the blob layout stays one tested thing.

**Defends against:** the file *leaving the machine* — a home-dir backup, an
`scp` of the data dir, a synced folder, a container image built over a live data
dir, a resold disk read on other hardware. The machine ID lives outside the data
directory and does not travel with a copy of it.

**Does not defend against**, stated plainly in the code and the whitepaper
rather than implied away:

- A local attacker running as the **same UID**. Whatever this process derives,
  they derive. That is why the OS keychain stays preferred wherever one exists.
- A **full-disk image or VM snapshot**. `/etc/machine-id` is world-readable and
  on the same disk. This beats selective exfiltration, not an image of
  everything.
- **Forensic recovery** of the pre-migration plaintext blocks — the rename
  unlinks the old inode, nothing securely erases it.

**The PIN is deliberately not in this KEK.** It cannot be: the keystore is read
before the PIN exists (boot reads `pin_meta_{uid}` to pick between the "enter
PIN" and "set PIN" screens). And it need not be: since #194 the actual secrets
in that file — `db_key_wrapped_{uid}`, `account_id_key_wrapped_{uid}` — are
already Argon2id(64 MiB, t=3) + XChaCha20-Poly1305 under a PIN-derived KEK.

The two layers are complementary and each covers the other's gap. The PIN layer
alone is weak against an exfiltrated file: 4 digits is 10⁴ candidates, about 40
minutes of single-core offline sweep at the tuned ~250 ms/guess. That is the
hole this closes. HKDF and not Argon2id here because a 128-bit machine ID has
nothing for a slow KDF to slow down.

**No TPM.** It would genuinely improve the first bullet — a sealed key is in no
backup and no disk image. Not integrated because `tss-esapi` is a heavy C
dependency on `tpm2-tss`, needs both a TPM and a reachable resource manager, and
needs three platform integrations. A half-integration with a silent fallback
would advertise protection a deployment might not have. Recorded as a deliberate
omission.

Hosts with no machine ID at all (some minimal containers ship an empty
`/etc/machine-id`) get a loud error naming `POLLIS_KEYSTORE_MACHINE_ID`, not a
constant-key fallback and not a plaintext one.

## Migration safety

Losing an identity key is unrecoverable, so every branch here favours "keys
survive" over "plaintext is gone one session sooner":

- A pre-#882 plaintext file stays readable. The first read *or* write
  re-encrypts it in place through the existing tempfile + fsync + rename.
- Encoding completes **before** the old file is touched, so every interruption
  leaves either the complete old file or the complete new one — never a mixture.
  Test: `interrupted_migration_keeps_every_key` reproduces the exact crash state
  (partial `.json.tmp` next to an intact original).
- A migration that fails does **not** fail the read
  (`a_failed_migration_does_not_fail_the_read`).
- A file that fails to *decrypt* is never rewritten, never backed away, never
  treated as empty (`undecryptable_file_is_left_on_disk_untouched`). The
  pre-existing back-up-on-corrupt path can now only fire for JSON that is
  garbage *after* a successful decrypt.
- No plaintext fallback on a write that cannot seal
  (`a_write_that_cannot_seal_leaves_the_old_file_intact`).

The filename stays `dev-keystore.json` on purpose: renaming means writing a new
file and then deleting the old one, and that delete is a delete of live key
material. Keeping the name lets the atomic rename overwrite the plaintext in
place.

## Backend selection

`BackendKind` has exactly two variants, `OsKeychain` and `EncryptedFile`. The
plaintext third state is gone from the type, not just from the code path — the
cliff was that it was reachable by flipping one Cargo feature. The choice is
made once per process by a read-only probe and then frozen: a keychain that
answers on one call and errors on the next must not split this process's slots
across two stores (#184's failure class).

## Guards

- `guard_written_bytes_are_never_plaintext` — the store file must not contain
  the slot name or the value and must not parse as the plaintext JSON store.
- `pollis-core/tests/keystore_at_rest.rs` — the shipped `OsKeystore` against a
  real `POLLIS_DATA_DIR`, inspecting the bytes that land on disk.
- `guard_cli_keeps_the_os_keychain_backend` (pollis-tui) — fails the TUI's test
  run if `os-keystore` is dropped from its dependency again.
- `guard_selected_backend_is_never_plaintext`,
  `guard_backend_choice_is_stable_within_a_process`,
  `debug_builds_use_the_file_backend`.

Every security-relevant test was revert-verified — the code was broken one way
at a time and each test confirmed to fail.

## Verified live, release build

```
with secret-service:    OS keychain, no keystore file written
without (DBUS broken):  [keystore] OS keychain unavailable (…No such file or directory)
                        [keystore] no usable OS keychain on this host; falling back
                                   to the machine-bound encrypted file store
                        store → load round trip OK, file bytes sealed
```


actuallydan:
## Post-merge release checklist (`docs/deployments.md`)

Merged as `4239d1b0`. Blast radius walked; decision recorded for every output.

**Touched:** `pollis-core/` (`keystore.rs` + one new test file), `pollis-tui/`, docs.
**No wire contract changed** — no canonical signing, `livekit_jwt`, `r2` presign, DS
request/response shape or MLS envelope encoding. This is purely local at-rest storage.

| Output | Decision |
|---|---|
| **CLI (`pollis`)** | **— redeploy (owed).** See below. |
| **Desktop app** | **— defer.** On a release desktop the keychain probe answers and the path is unchanged; the fallback only engages where no keychain exists. Rides the next `v*` tag. |
| **Mobile (uniffi)** | **— N/A.** `android_kek` / `ios_kek` are untouched; the change only refactors them behind `FileCodec`. `android-check`, `ios-check` and `android-build` all pass on this SHA. |
| **DS (`pollis-delivery`)** | **— N/A.** No wire contract touched and `pollis-delivery/` not modified. |
| **Relay pool** | **— N/A.** `pollis-relay` / `pollis-device-cert` untouched. |
| **Verifier + transparency log** | **— N/A.** `verifiable-log*` untouched. |
| **DB migrations** | **— N/A.** None in this change. |
| **Website / LiveKit** | **— N/A.** |

### Why the CLI redeploy is owed, not optional

The currently-shipped CLI was built **2026-08-16 21:12** from `2de16f34`, which
contains both #879 and #881. That build has `os-keystore` on with the *old*
compile-time backend selection — so on a box with no secret-service (a server over
SSH, the CLI's primary home) `keyring` fails at runtime and the terminal client
**cannot start**. That is the live regression #882 exists to clear.

`cli-release.yml` is `workflow_dispatch`; the last successful run resolved version
`1.1.0` from the workspace `Cargo.toml`, so a plain dispatch off `main` with default
inputs matches the convention — no version bump needed. It publishes to R2 under the
`cli/` prefix plus cosign provenance, and touches no AUR path, so the ongoing AUR
gateway outage does not block it.

After dispatching, confirm the artifact actually republished (checklist step 2 —
"confirm the new build is live, not merely that a workflow fired").

PR #954, opened 2026-08-17
Explain v1.9.3, and make the rebuilder say which kind of red it found (#944)

### LINKED ISSUE #944: v1.9.3 did not reproduce, and it has never been explained
Found while publishing the rebuild-verify history for #877.

#939 fixed a false-red bug in `rebuild-verify.yml` (the wait loop broke on the first attempt, because `pollis-verify release` exits 0 with an empty artifact list for a not-yet-published tag). It was tempting to write off the red runs as that bug — but classifying each run individually against its log shows **two were false and one was real**:

- **v1.9.4** — false red (the loop bug)
- **v1.9.5** — false red (the loop bug; it rebuilt byte-identical, 0 differing bytes)
- **v1.9.3 — a genuine non-reproduction, still unexplained**

Reproducible builds are one of the strongest claims on the site, and #937 has just rewritten that copy to say the Linux AppImage reproduces. One unexplained mismatch in the published history is exactly what a skeptical reader will find and ask about, so it is better to have the answer than to hope nobody diffs the ledger.

Worth noting the related known-unknown: #750 established that rustc itself is non-deterministic in ways no compiler profile fully avoids, and macOS builds are normalised for signatures. So the answer may turn out to be benign — but "probably benign" is not what belongs next to a published verdict.

Investigate what differed, and either explain it in `website/rebuild-ledger.json` alongside the verdict or fix the cause.

actuallydan:
Closes #944.

## v1.9.3 reproduces

Re-running the identical check on 2026-08-17 landed on `ubuntu22@20260810.260.1` — the exact runner image the v1.9.3 release built on — and reproduced `07bccca82cd4b16a57899bc2f18b81ba69595984d2c417b36289a9be278eebd3`, **byte-identical** to the logged payload ([run 32036282718](https://github.com/actuallydan/pollis/actions/runs/32036282718)).

Same tag, same public source, same recipe. One variable differed in August; holding it constant makes the bytes match. **Benign — the shipped binary corresponds to the tagged source.**

## What differed in August

The August 13 rebuild was a real byte divergence, and it is fully accounted for.

**Dominant cause: the two builds ran on different CI runner images.**

| | |
|---|---|
| Release `build-linux` | `ubuntu22@20260810.260.1` |
| Rebuild | `ubuntu22@20260720.234.2` |
| Gap | 36 minutes |

GitHub rolls a re-imaged `ubuntu-22.04` out gradually, so the rebuild landed on the *older* image. `linuxdeploy` copies the app's system shared libraries off the build runner into the AppImage, so seven vendored libraries differed in real code — `libgssapi_krb5`, `libk5crypto`, `libkrb5`, `libkrb5support`, `libsqlite3`, `libsystemd`, `libudev` — changed `.text` sizes and shifted symbols, not just build ids (`libsqlite3.so.0` `.text`: `0xeaf9d` → `0xeaf3d`). Residual §6, measured rather than assumed.

The published "105,863,272 differing bytes" is compression smear across a squashfs, not the size of the change.

**Only other difference, in `usr/bin/pollis`:** the *order* of the twelve CSP `sha256-` hashes in `.rodata` — same twelve values, permuted. `tauri-codegen` pushes one per `.js` file in `WalkDir`/`readdir` order and never sorts (`embedded_assets.rs`, `RawEmbeddedAssets::new`). That the digests are identical is itself proof the frontend bundle reproduced byte-for-byte. Not fixed here on purpose — the three available fixes (fork `tauri-codegen` behind `[patch.crates-io]`, weaken CSP via `dangerousDisableAssetCspModification`, or collapse the frontend to one chunk) are each worse than a permutation of an unordered allow-list. Documented as §6a.

## What changed

The comparison was wrong, so that is what got fixed. "Did not reproduce" is the strongest accusation this CI can make, and the hash comparison alone cannot know whether it is entitled to make it.

- **`ReleaseArtifact` now carries the leaf's `toolchain`**, so `pollis-verify release --json` exposes the recorded recipe. Read through the *verified* report — a leaf whose inclusion proof checked out under the pinned key — because this is the one field that can talk the verdict out of an accusation, and an unauthenticated fetch would let whoever serves verify.pollis.com fake an image mismatch to suppress a real finding.
- **`rebuild-verify.yml` classifies four outcomes** instead of two: `reproduced`, `did_not_reproduce` (same image, differing bytes — the real finding), `environment_drift` (images differ — inconclusive), `recipe_unrecorded` (leaf predates #939 and says `unknown`). **All three non-green outcomes still fail the run** — an inconclusive rebuild is not a passing one, and auto-greening drift would open a window for a genuine divergence every time GitHub re-images a runner. Only the wording, title and summary change.
- **The failure diagnostic splits differing paths by origin** — `usr/lib/*` is vendored off the runner, everything else is ours — and says so explicitly when only vendored libraries differ. That single line is what the last investigation lacked.
- **The ledger stopped publishing feature-branch development runs as release verdicts.** Found while working here: 14 runs from `fix/build-recipe-drift`, `diag/rebuild-artifact` and `feature/auto-verify-releases`, 13 of them red, were dispatched against real tags and so resolved a tag and got published against v1.8.4. The existing "no tag = development run" rule only caught the weaker half.
- **Ledger, `/assurance`, `/security`, `/faq`, `/learn`** no longer say v1.9.3 is a real, unexplained non-reproduction; they say what it was, and that it is not a finding against the shipped binary.

Also documented: the AppImage icon symlink is picked from an unordered listing (§6b), and Tauri fetches two of its five AppImage tools from `master`/`continuous` refs that nothing pins or records (§6c) — the bundler hardcodes the URLs, so the only lever is `TAURI_BUNDLER_TOOLS_GITHUB_MIRROR_TEMPLATE`, which cannot reach the two `raw.githubusercontent.com` scripts.

## Verification

Both directions, because a fix that makes everything pass would be worse than the bug.

- `scripts/test-rebuild-verdict.sh` — **22 passed, 0 failed**. Asserts `did_not_reproduce` still fires on same-image divergence and is still flagged as an accusation; that drift and unrecorded recipes are not; that a helper-image-only drift still counts as drift; that a hash match wins regardless of environment; and that an empty rebuild hash never reads as reproduced (the #939 vacuous-truth failure, one layer down). Gated by `scripts-check.yml`.
- Replayed against v1.9.3's real numbers (reproduced `a6767a4c…`, logged `07bccca8…`): as the leaf actually is → `recipe_unrecorded`; with the two real image ids → `environment_drift`; same image → `did_not_reproduce`; hashes equal → `reproduced`.
- `cargo test -p verifiable-log-serve` — **65 passed, 0 failed**, including a new test that the recipe survives into both the struct and the serialized JSON a rebuilder reads.
- `cargo test -p pollis-core --lib transparency` — **50 passed, 0 failed** (fixtures now carry a recipe, since a report without one is not one the log could serve).
- `scripts/test-attest-helpers.sh` — **100 passed, 0 failed**. `check-workflows.py`, `check-build-recipe.py` clean. `tsc` clean.
- The new workflow was dispatched end-to-end against v1.9.3 from this branch, not just unit-tested — **which is how a bug in this PR was caught before merge.**

  Every checkout in `rebuild-verify.yml` is `ref: $TAG`, so `scripts/` in the workspace is the *released tag's* copy. Sourcing the new verdict library from there fails on every tag older than it, and the first symptom was silent: `runner_image_id: command not found` → empty command substitution → `echo` succeeds → job green with no image recorded, degrading every later verdict to "inconclusive" while looking fine. The rebuilder's own tooling now comes from a second checkout at the workflow's own ref (`.rebuild-tools`), placed after every build step so it cannot be an input to the bytes being judged. `payload-hash.sh` deliberately still comes from the tag, since that one must hash the payload the way the release did. `test-rebuild-verdict.sh` asserts all four of those wiring facts — verified to fail (exit 1) when the bad form is reintroduced and pass (exit 0) when it is not.

**The shipped workflow step itself was executed in all four directions**, by extracting the `run:` body of `Assert reproduced hash is the logged Linux payload` straight out of the YAML and running it against a real `release-report.json` with the `.rebuild-tools` layout in place:

| Case | Verdict | Exit | Title |
|---|---|---|---|
| hashes match, same image | `reproduced` | **0** | ✅ Reproduced |
| hashes differ, same image | `did_not_reproduce` | **1** | Did not reproduce |
| hashes differ, different image | `environment_drift` | **1** | Inconclusive — build environment drifted |
| hashes differ, no recorded recipe | `recipe_unrecorded` | **1** | Inconclusive — the release recorded no build recipe |

Both failure step-summaries were checked as rendered markdown, not just as code.

Two attempts to run the new workflow end-to-end on a real runner both died in the Tauri bundler with `failed to bundle project: http status: 429` — GitHub rate-limiting the AppImage tooling downloads described in §6c, after a 14-minute compile. That is the residual itself biting, not a defect in this change, and it is now recorded in the doc with the evidence. The parts that need a real runner are separately proven: `.rebuild-tools` + `runner_image_id` produced `ubuntu24@20260810.271.1` in a live `rebuild-capture-helper` job, and the full build→hash→compare path is what produced the green `REPRODUCED` on run 32036282718.


actuallydan:
Landed on `main` as 79193a82 (squash). GitHub's API was returning 503s during the merge, so the PR record was never updated even though the merge itself went through — every one of the 18 changed files is byte-identical on `main`. Closing manually.

Issue #947, opened 2026-08-16
Device revocation writes no security_event and notifies nobody


actuallydan:
Found during #877's publishing phase and published as a stated gap on the new `/device-security` page.

When a device is revoked:
- no `security_event` row is written, so the account's own audit log has no record of it
- no contact is notified, so peers who have verified that device learn nothing

The recovery machinery exists and works — this is about the trail it leaves. For a "my laptop was stolen" flow, the user's own security log is the first place they will look to confirm the revocation took effect, and a peer who verified the stolen device has a legitimate interest in knowing.

The page now tells users this honestly rather than implying a notification they will not receive, which is the right interim state — but the gap is worth closing, and the `security_event` half looks small.

actuallydan:
Fixed in #958 (`8d338702`). `revoke_device` posts a `device_revoked` event through the existing `POST /v1/security-events`, carrying the device id and its name — the `user_device` row is tombstoned, so the event is the last place that name survives — written before the reconcile loop so the trail does not depend on N commits. Peer notification is deliberately not done and the reason is recorded on the PR: the `device_removed` banner surface can never fire, because its only producer is the actor's own reconcile and §5.3 strips device ids from the cleartext broadcast. Closing it means deriving the diff from the authenticated commit inside each peer's commit-processing path; a server-visible column would let a hostile DS fabricate the banner.

Issue #927, opened 2026-08-16
Message history is capped at 50 per conversation — next_cursor is threaded but never consumed


actuallydan:
Found during #874's IPC phase.

`useMessages` fetches 50 messages and the response's `next_cursor` is returned and threaded through the TypeScript type, but **nothing ever pages backwards**. So the UI silently caps every conversation's visible history at its most recent 50 messages, regardless of what the local DB holds.

This matters more here than in most products: CLAUDE.md's "Messages must work; history is bounded, not flaky" section enumerates exactly three acceptable losses, and "the UI only renders the last 50" is not one of them. The messages are on disk and decryptable — they just cannot be reached.

Note #874's render phase is separately looking at virtualising the message list and bounding `olderMessages`, which `loadMore` currently grows without limit. Whatever paging is added here should land on top of that, not fight it.

actuallydan:
Fixed in #958 (`8d338702`). Two defects, not one. The filed one: nothing consumed `next_cursor`. The worse one, found while fixing it: the paging cursor was seeded by an effect that read the *previous* conversation's pages, so reopening a conversation whose first page was still cached pinned it to its newest 50 permanently, with no dependency that could ever change to make it retry — three clicks to reproduce. `MainContent` now derives pages and cursor during render from state stamped with the conversation. Also fixed: load-more only fired on a scroll event, so at offset 0 the browser had no scroll left to emit and the log dead-ended.

Issue #911, opened 2026-08-16
InviteLinkRow uppercases already-translated copy with a locale-insensitive .toUpperCase()


actuallydan:
Found during #902, left alone as out of that ticket's scope.

`InviteLinkRow` calls `.toUpperCase()` on status copy that has already been through `t()`. Locale-insensitive casing on translated text is an i18n smell: `.toUpperCase()` without a locale uses the invariant mapping, which is wrong for Turkish dotted/dotless i, among others.

Correct for all seven shipped locales today (none is Turkish), so this is not currently user-visible — but it becomes a real bug the moment `tr` is added, and it will not be obvious then that a casing call in an invite row is the cause.

Fix: either `toLocaleUpperCase(i18n.language)`, or drop the casing and let the catalogue supply the intended form (preferable — some scripts have no case at all, and the Arabic and Chinese catalogues are already unaffected by it in a way that makes the transform meaningless there).

actuallydan:
Fixed in #958 (`8d338702`). `localeUpperCase(status, activeLocale())`, via a new helper in `i18n/languages.ts`. The transform was kept rather than moved into the catalogue: no shipped locale cases differently today, so only a unit test can see the bug, and adding terminal-skin keys would have left six locales rendering English for a badge they already translate correctly.

Issue #898, opened 2026-08-15
Invite-link copy swallows failures and bypasses the clipboard bridge


actuallydan:
`CreatedInviteLinkCard.handleCopy` has the same defect #889 fixed for message permalinks: a failed clipboard write is caught and only `console.error`d, so it looks exactly like a success — the button flips to "Copied" only on the success path, but the user is told nothing when it fails.

It also calls `navigator.clipboard.writeText` directly. The bridge has `writeClipboardText`, which is backed by a Rust command specifically because `navigator.clipboard` is unreliable on WebKitGTK — the Linux webview this app ships on. So on the platform most likely to fail, this path both uses the unreliable API and hides the failure.

Fix: route through `writeClipboardText` and surface the false return, matching the three-state treatment now on the message copy-link button.

Found while fixing #889.

actuallydan:
Fixed in #958 (`8d338702`). Routed through the bridge's `writeClipboardText` — the Rust-backed path that exists because `navigator.clipboard` is unreliable on WebKitGTK — with the same three-state treatment #889 gave message permalinks: `data-copy-state` plus the accessible name, so a failed copy can no longer look identical to a successful one.

PR #958, opened 2026-08-17
Fix four user-visible bugs: unreachable history, silent revocation, hidden copy failures, locale-blind casing

actuallydan:
Four user-visible bugs, one PR, a commit each.

### #927 — history was capped at 50 per conversation
Two defects, both of which end with decryptable messages on disk that the UI cannot reach.

1. **The paging cursor was seeded from an effect that read the previous conversation's pages.** Opening a conversation whose first page was still in the React Query cache found `olderMessages` non-empty (it belonged to the conversation just left), declined to seed a cursor, and had no dependency left that could ever change to make it try again. That conversation stayed pinned to its newest 50 for as long as the cache held it — reproducible in three clicks: read back in A, switch to B, B is capped. `MainContent` now derives the pages AND the cursor during render from state stamped with the conversation they belong to; the reset effect and the seeding effect are both gone.
2. **The load-more trigger was a scroll event only.** At scroll offset 0 the browser has no scroll left to emit, so a prepend that lands the reader back at the top dead-ends the log on the page it has. `MessageList` re-checks when the timeline changes as well; the conditions inside the check (still at the top, or the log does not fill its own viewport) are what stop it running away — the ordinary prepend re-anchors the reader well below the threshold and it fetches exactly one page.

Built on #874's windowing rather than against it: an off-window message is one array entry and no DOM.

### #947 — device revocation left no trail
`revoke_device` now appends a `device_revoked` row to the account's own security log via the existing `POST /v1/security-events`, carrying the device id and its name (the `user_device` row is tombstoned by the revocation, so the event is the last place the name can be read). Written immediately after the tombstone lands and before the per-conversation MLS reconcile loop, so "did it take effect?" does not depend on N commits succeeding. Best-effort, like the enrollment events beside it: the revocation has already committed, and failing the command on a flaky audit write would report a failure for something that happened. `SecurityPage` renders it as a sentence; `website/device-security.html` no longer tells users it is not recorded.

**What remains, and why it is not in this PR.** Peer notification is a design decision, not a gap. Safety numbers in Pollis identify a *person* — every device chains to one account identity — so revoking a device changes nobody's safety number, which the public page already explains. The in-app surface that could carry it (`device_removed` roster banners) exists but can never fire today: its only producer is the acting client's own reconcile, and §5.3 metadata minimisation deliberately strips device ids from the cleartext LiveKit broadcast, so no peer ever receives a non-empty `devices_removed`. Closing it honestly means deriving the diff from the authenticated commit inside each peer's own commit-processing path — the derivation `reconcile.rs` already promises in a comment — which is an MLS-core change of its own size, and inventing a server-visible "removed" column instead would hand a hostile DS the ability to fabricate the banner.

### #898 — invite-link copy hid its failures
`CreatedInviteLinkCard` routed through `navigator.clipboard` (unreliable on WebKitGTK, the Linux webview we ship) and `console.error`d the failure, so a failed copy of a link that can only ever be shown once looked exactly like a success. Now goes through the bridge's `writeClipboardText` and surfaces the false return as the same three states #889 gave the message copy-link button — `data-copy-state` plus the accessible name.

### #911 — locale-insensitive casing of translated copy
`InviteLinkRow` upcased status copy that had already been through `t()`. Now `localeUpperCase(status, activeLocale())`, a new helper in `i18n/languages.ts`. Kept the transform rather than moving the form into the catalogue, on testability: the bug is invisible in all seven shipped locales, so only a unit test over the helper can ask the question, and pushing three terminal-skin variants into `en` would have left six locales rendering English for a badge they translate correctly today.

### Also
- `frontend-check.yml` now runs the renderer's unit tests. Nothing did, which is why #911's guard would otherwise have guarded nothing.
- `rtl.spec.ts`'s arrow-key test was flaky on `main` (one full-suite run in four): a render landing between `focus()` and `press()` spends the keystroke on the body, and retrying the assertion cannot get it back. The focus/press/assert is now one retried step. Verified it still fails with `arrowStep`'s RTL handling reverted.

### Verification (raw)
| | before | after |
|---|---|---|
| Playwright e2e | 214 passed | **222 passed** (2 consecutive clean full runs) |
| `cargo test -p pollis-core --no-default-features --features test-harness` | 614 passed | **614 passed** |
| flows (`revoking_a_device_records_a_security_event`) | — | **1 passed** |
| `frontend` unit tests | 30 passed | **33 passed** |
| `tsc`, `pnpm build`, `node scripts/i18n-check.mjs` | — | clean / clean / OK (untranslated warnings only) |

Every new test was re-run with its fix reverted:
- reopened-from-cache paging → **fails** both skins (`toBe(true)` on reaching the oldest message)
- revocation security event → **fails** (`got []`)
- invite copy failure + confirmation → **fails** both skins (no `data-copy-state`)
- Turkish casing → **fails** (`'AKTIF' !== 'AKTİF'`)
- "scrolling back reaches the oldest message" passes on `main` — labelled **(guard)** in the test name; the fresh-visit wheel path already worked, and it pins that the cascade fix did not break it.


actuallydan:
Post-merge release checklist (`docs/deployments.md`):

Blast radius: `frontend/` + `pollis-core/src/commands/auth.rs` (one added call to the **existing** `POST /v1/security-events` with the existing body shape — no wire-contract change, so the DS's hand-written mirror cannot drift) + `website/` + one CI workflow.

- **Website** — **redeployed** (`website-deploy.yml` run 32044565347); `device-security.html` no longer tells users revocation is unrecorded. Confirmed live by polling the URL, not by the workflow going green.
- **Desktop app** — **defer**: all four fixes are user-visible and want a release, but cutting a `v*` tag is a deliberate act and not part of this ticket. The next tag picks them up; no migration is pending behind it.
- **CLI (`pollis-tui`)** — **defer**: shares `pollis-core`, so a revocation from the TUI now writes the audit row too, but the change is additive and the CLI renders no security log.
- **Mobile** — **defer**: same additive `pollis-core` change, no mobile surface touched.
- **DS** — **N/A**: no `pollis-delivery` change and no mirrored contract.
- **Relay / verifier / transparency log / LiveKit** — **N/A**.
- **Migrations** — **none**.

Issue #934, opened 2026-08-16
MainContent's load-more trigger uses a hardcoded 150px and clears its flag in the prepend batch


actuallydan:
Two small pre-existing items in `MainContent`, noticed during #874's virtualisation phase and left untouched there to keep that diff on the windowing change.

1. **The load-more trigger is `container.scrollTop < 150`** — a hardcoded pixel constant in a codebase where sizes that track the user's font setting are supposed to be in `rem` (CLAUDE.md). At a large font setting, 150px is a much smaller fraction of a row than it is at the default, so the trigger fires later relative to what the user sees.
2. **The load-more `finally` sets `loadingMore(false)` in the same batch as the prepend**, so there is no frame in which the new rows exist and the flag is still set. Harmless today; it removes the seam anything wanting to react between those two states would need.

Neither is user-visible now. Filed so they are written down rather than rediscovered by whoever next touches scrolling — #874 added windowing on top of both.

actuallydan:
Fixed in #960 (`d7e869cc`).

Issue #933, opened 2026-08-16
.codesight/wiki/ui.md's inventory is stale and has no generator


actuallydan:
Found during #874's render phase: `ui.md` claimed 111 `.tsx` files; there are **147**. The article says to regenerate rather than patch, but no generator exists in the repo — so the instruction cannot be followed and the count has drifted by a third.

#874 corrected the number and the sections it touched, but a hand-patched inventory will drift again by definition.

Fix: either write the generator the article assumes, or drop the exhaustive inventory and keep the parts that are actually maintained by hand. A doc that tells you to regenerate it with a tool that does not exist is worse than one that admits it is partial — the same problem #876 was closed for.

actuallydan:
Fixed in #960 (`d7e869cc`).

Issue #932, opened 2026-08-16
Sidebar settings rows have no data-testid, so tests locate them by translated label


actuallydan:
Found during #874's render phase.

`SectionHeader` emits `sidebar-row-*` testids, but the sidebar's settings rows (Preferences, Security, Voice Settings, …) carry none. Every e2e test that navigates to them therefore matches on the **translated** label.

That was harmless when the app was English-only. Since #855 it means a locale change can break navigation tests for reasons that have nothing to do with navigation — and the failure will look like a routing bug rather than a selector bug.

Fix: give them stable testids like their siblings, and switch the specs that navigate by label.

actuallydan:
Fixed in #960 (`d7e869cc`).

Issue #930, opened 2026-08-16
enforce_cache_cap_now stats the entire media cache on every window focus


actuallydan:
Found during #874's IPC phase.

`lib.rs` wires `enforce_cache_cap_now()` to app focus. #874 removed every other focus-triggered cost (7 IPC calls → 0), so this is now the only work that happens when you alt-tab back — and it walks the whole media-cache directory to total its size.

Cheap today, but it scales with the cache rather than with anything the user did, and the cache is exactly what grows over months of use. A user with a large media cache pays for it on every focus.

Fix options: maintain a running total updated on write/evict, or run the sweep on a schedule tied to cache writes rather than to focus. Note CLAUDE.md bans periodic polling, so a timer is not the answer — event-driven on cache mutation is.

actuallydan:
Fixed in #960 (`d7e869cc`).

PR #960, opened 2026-08-17
Six frontend cleanups: #928, #929, #930, #932, #933, #934

actuallydan:
Six unrelated frontend cleanups, sharing a PR for speed. **One commit per issue** — that boundary is the only thing making them separable in review or individually revertable, so please don't squash.

| commit | issue |
|---|---|
| `fix(frontend)` | #928 — `useUserProfile`'s `queryFn` wrote to a MobX store |
| `chore(frontend)` | #929 — dead hooks, Clerk-era api stubs, half-removed phone field |
| `perf(media-cache)` | #930 — `enforce_cache_cap_now()` walked the cache on every window focus |
| `test(sidebar)` | #932 — settings rows had no testid, so specs navigated by translated label |
| `fix(message-log)` | #934 — px load-more threshold; flag cleared in the prepend batch |
| `docs(ui)` | #933 — `ui.md` told you to regenerate with a tool that did not exist |

## What changed

**#928** — the store write moved into a `useEffect` keyed on the resolved value. A `queryFn` runs on refetch, retry and cache-restore with no ordering guarantee, and is skipped entirely on a cache hit — so the old code was both mutated by background refetches and *missed* exactly when the data was already cached. `VoiceSessionManager` still reads `appStore.userAvatarUrl` for the local voice tile, so the write itself had to stay.

**#929** — deleted `useGroupChannels` (zero callers; `groupQueryKeys.channels` stays, it is still an invalidation target), `uploadFile`/`downloadFile` and the five Clerk-era stubs including `cancelAuth` (also uncalled, not named in the issue). The phone field is gone from `SettingsPage` along with its state, its save wiring, its `ServiceUserData` field and the `useUpdateProfile` argument. `api.updateUserProfile` keeps its `phone` parameter — it mirrors the Rust command, which mobile still uses.

**#930** — the focus hook is gone; the cap is enforced on cache *writes*, which every write path already did. A cache nobody writes to cannot grow past it, so external tampering is still caught at the next write. `enforce_cache_cap_now()` is deleted rather than left public, since it existed only as the focus hook's entry point.

**#932** — `Row` takes an optional `testId` and emits `sidebar-row-*`, the scheme `SectionHeader` already used. `render-cost.spec.ts` and `right-panel-persistence.spec.ts` switched off labels. Also added the two missing page testids (`voice-settings-page`, `update-page`) the new assertions needed.

**#934** — `loadMoreThresholdPx()` in `messageWindow.ts`, 10rem, read at call time so it follows a font-size change under a mounted log; 10rem is exactly the old 150px at the 15px default, so nothing moves for a default user. The flag now clears in a `useEffect` keyed on `olderPages` (only non-prepend paths clear inline), which puts the clear one commit after the rows rather than in the same batch. Both #927's paging and #874's virtualisation were read first; neither is regressed — their existing specs pass unchanged.

**#933** — wrote the generator the article assumed: `node scripts/ui-inventory.mjs`, `--check` to fail on drift, wired into the renderer's unit tests. Hand-written per-component notes survive regeneration (anything after the backticked path is preserved verbatim). The regeneration found `components/Emoji` (8 files) missing entirely and fixed two bad extractions the previous tool had left behind — `ALL_ALGORITHMS` for `DotMatrix.tsx`, and a sub-component's props on `VoiceSettingsPage`. The RightPanel narrative moved above the generated block, since prose cannot live inside it.

## Testing

Every behavioural fix has a test that fails without it; guards are labelled as guards in both the test names and `testing.md`. Per-fix revert results and raw before/after numbers are in the PR discussion.


actuallydan:
## Test results

### Raw numbers, before → after

| gate | before | after |
|---|---|---|
| Playwright (`e2e:ui`) | 222 passed | **227 passed** (+3 i18n/#932, +2 message-window/#934) |
| Renderer unit (`pnpm --filter frontend test`) | 33 passed | **44 passed** (+11) |
| `cargo test -p pollis-core --lib` | 619 passed, 1 ignored | **623 passed, 1 ignored** (+5, #930) |
| `tsc --noEmit` | clean | clean |
| `pnpm --filter frontend build` | clean | clean |
| `node scripts/i18n-check.mjs` | OK | OK |
| `node scripts/ui-inventory.mjs --check` | n/a (no generator) | OK |

Each of the six commits was also checked out individually: `tsc` clean and unit tests green at every one, so any single fix can be reverted on its own.

### Per-fix revert results

Each fix was reverted in isolation and the suite re-run.

| # | what was reverted | result |
|---|---|---|
| **928** | `appStore.setUserAvatarUrl(...)` put back inside the `queryFn` | 2 failed — `GUARD: no queryFn writes to a MobX store`, `useUserProfile hydrates the store from an effect` |
| **929** | `downloadFile` restored to `api.ts` | 1 failed — `GUARD: the Clerk-era and unused api.ts exports stay deleted` |
| **929** | `phone` state + `settings-phone-input` testid restored | 1 failed — `GUARD: the settings phone field leaves nothing behind` |
| **930** | `enforce_cache_cap_now()` restored and re-wired to `Focused(true)` | 2 failed — `guard_window_focus_does_not_sweep_the_media_cache`, `guard_no_public_entry_point_for_a_focus_time_sweep` |
| **932** | `data-testid` removed from sidebar `Row` | **7 failed** across all three specs — 3 in `i18n.spec.ts`, 2 in `render-cost.spec.ts`, 2 in `right-panel-persistence.spec.ts` |
| **933** | inventory count hand-patched back to 111 | 2 failed — `the ui.md component inventory matches frontend/src`, `the inventory covers every .tsx exactly once` |
| **934** | `LOAD_MORE_THRESHOLD_PX = 150` back in `MessageList` | 3 failed — both scaling tests + `GUARD: MessageList carries no px load-more constant` |
| **934** | `setLoadingMore(false)` back in the prepend batch | 2 failed (both skins) — `no batch contained the prepended rows while the log still reported fetching` |

### Which are guards, and why

Labelled as such in the test names and in `.codesight/wiki/testing.md`:

- **#928** — source scan over every `queryFn` in `hooks/queries/`. The symptom (a background refetch mutating a store out of order with render) is a race no deterministic test can provoke, so this follows the precedent `no-periodic-polling.test.ts` sets for rules *about* the code.
- **#929** — "these exports stay deleted". Not observable at runtime at all.
- **#930** — the two Rust guards. Paired with three behavioural tests: eviction is oldest-first, stops at target, and ignores `.tmp` files. `cache_dir_walks()` (test builds only) instruments the directory walk, and `the_walk_counter_counts_one_walk_per_sweep` proves the counter counts — so "0 walks on focus" cannot pass on a broken counter.
- **#933** — `--check` drift detection, paired with a round-trip test that hand-written notes survive regeneration.
- **#934** — the "no px constant" scan. Paired with the real arithmetic (10rem = 150px at the 15px default; the threshold is a constant multiple of the root at every size) and the browser-level seam test.

**#932 is not a guard** — the three new `i18n.spec.ts` tests drive a synthetic locale that renames every settings row, assert the labels really changed, and then navigate all seven rows by testid. That is the actual failure mode the issue describes, not a proxy for it.

### On not regressing #927 and #874

Both were read before touching scroll behaviour. Their specs pass unchanged: `older pages prepend without moving the message being read`, `scrolling back reaches the oldest message`, `a conversation reopened from cache can still reach its oldest message` (#927), and the whole windowing/render-cost set (#874). The `loadingMore` change deliberately clears the flag in a passive effect, which runs *after* the virtualiser's layout-effect re-anchor — so the anchoring #874 relies on is untouched.

actuallydan:
Amended the #929 commit (force-push, `7a520195` → `95b58c2a`): `.codesight/wiki/libraries.md` still listed `useGroupChannels` among `useGroups.ts`'s exports. Only change vs the previously-pushed head; commit structure and every test result above are unchanged.

Deliberately **not** touched while looking for other stale references:
- `.codesight/CODESIGHT.md` / `.codesight/libs.md` — tool-generated, "last scanned 2026-04-13", already months stale. Hand-patching a generated file is the exact anti-pattern #933 is about, and no generator for them is in the repo.
- `frontend/docs/frontend-audit.md` — carries a "SUPERSEDED — historical snapshot" banner. Editing a dated record to match today's code would falsify it.

Issue #951, opened 2026-08-17
Legacy unwrapped key slots still hold raw private keys for pre-PIN upgraders


actuallydan:
Noticed during #882.

`db_key_{uid}` and `account_id_key_{uid}` — the pre-PIN, *unwrapped* slots — still exist for users who upgraded from before #194, with `account_identity.rs` falling back to them. They hold raw private key material that was never PIN-wrapped.

#882 improved their situation incidentally: on the file backend they now sit inside the file-level ciphertext, so they are no longer plaintext on disk. But they are still not protected by the PIN-derived Argon2id layer that #194 introduced for everyone else, which means an attacker who defeats the file layer gets them directly rather than facing a slow KDF.

This is #194's migration rather than a new problem — the cutover for those users was never completed. Worth finishing: read the legacy slot, wrap it under the PIN-derived KEK, write the wrapped slot, and only then remove the unwrapped one.

actuallydan:
Fixed in #977. `set_pin` now writes the wrapped slots, **reads them back and unwraps them**, and only then deletes the legacy unwrapped originals — a `store` returning `Ok(())` is a claim, not evidence, and the failure is survivable while the legacy slot exists but unrecoverable the moment it does not. Revert-verified: removing the read-back fails `an_interrupted_migration_loses_nothing`.

Issue #950, opened 2026-08-17
Keystore residue from #882: misleading filename, backend-flip UX, and a corrupt-file key-loss path


actuallydan:
Three items noticed during #882, each deliberately left out of that PR.

1. **`dev-keystore.json` is now a misleading name.** After #882 it is the *production* store on headless installs, and it is no longer JSON — it is AES-GCM ciphertext. Renaming was rejected on purpose: a rename means deleting the old file, and that delete is a delete of live key material. Worth doing as its own change with a careful two-step (write new, verify readable, then unlink).

2. **A box that later gains a secret-service flips backend and appears empty.** The keychain becomes available, selection prefers it, the store is empty, and the user gets PIN-setup / OTP. Nothing is destroyed and it self-heals, and it is strictly better than today's hard failure — but it is a confusing experience. A "stay on the file if it holds current-namespace keys" rule would fix it and adds its own failure modes, so it needs designing rather than patching.

3. **The corrupt-JSON path still backs the file away and effectively starts over.** Pre-existing, and #882 narrowed it (it is now only reachable *after* a successful decrypt, so it cannot fire on ciphertext), but it remains a key-loss path for genuinely corrupt content. Losing an identity key is unrecoverable, so "start over" is a harsh default.

Also minor: `pollis-core/tests/keystore_at_rest.rs` uses `std::env::set_var`, which will need `unsafe` when the workspace moves to edition 2024.

actuallydan:
Fixed in #977. `dev-keystore.json` → `keystore.pks` via write-new → read-back → unlink-old; a host that later gains a Secret Service now stays on the file while it holds current-namespace keys; and a corrupt store is left in place rather than renamed aside — the old `.bad-{ts}` rename looked like a backup and behaved like a wipe. Revert-verified on both.

Issue #946, opened 2026-08-16
The whitepaper has three stale claims about client-side credentials


actuallydan:
Found during #877's publishing phase.

The whitepaper states that the client bakes in Resend, R2 and LiveKit credentials. **It no longer does** — those paths moved behind the Delivery Service. #877 published the *correct* version on the new `/assurance` and `/subprocessors` pages, but did not rewrite the whitepaper itself.

That leaves the same defect class #937 spent a whole phase fixing: a published document contradicting the code, on the single artifact most likely to be read by an auditor or investor doing diligence. It also understates the product — the current architecture is better than what the whitepaper describes.

Fix: reconcile the whitepaper against the code, using the newly published pages as the reference for what is now true.

actuallydan:
Fixed in #977. The whitepaper actually contradicted itself — §1 said the client-side credentials were gone while §§4/8/9/10/12/13 still said they were baked in. Reconciled against the code, and the same stale claim was live on `website/security.html` and is fixed there too.

Issue #945, opened 2026-08-16
The pinned ML-DSA key is hand-copied into eight places with no CI check that they agree


actuallydan:
Found during #877's publishing phase. The ticket said three places; the real count is **eight**.

The transparency log's ML-DSA-44 public key is duplicated by hand across the codebase, the website and the docs, and **nothing verifies the copies agree**. A rotation (or a typo) that updates seven of eight leaves one surface pinning a key the log no longer signs with — which, given #875's work, is precisely the shape that makes an honest log look like a tampered one to whichever client reads the stale copy.

#877 published this as a stated limit rather than papering over it, but it wants a real fix: one source of truth, or a CI check that fails when the copies diverge. `website-verify.yml` already asserts the pinned key in `artifacts.js`, so the pattern exists — it just needs to cover all eight.

A rotation has been executed before (#732), so this is not hypothetical.

actuallydan:
Fixed in #977. `transparency.rs` is now the single source of truth for the pinned ML-DSA key, with an offline check gating every PR that fails when a copy drifts **and** when a new unregistered copy appears — so the count cannot silently grow past eight. Proven to fail when a copy is changed.

PR #977, opened 2026-08-17
Five security and keystore fixes: #951, #950, #945, #946, #942

actuallydan:
Five independent security/keystore issues. One commit per issue — they share a PR
for speed, not because they are related, so the commit boundary is what makes
them separately reviewable and separately revertable.

### #951 — legacy unwrapped key slots
`set_pin`'s wrap-then-delete now runs through `pin::install_wrapped_slots`, which
**reads the three new slots back and unwraps them** before removing the legacy
`db_key_{uid}` / `account_id_key_{uid}`. Until that read-back succeeds those slots
are the only copy of an unrecoverable key on the device, and a `store` returning
`Ok(())` is a claim, not evidence — a keychain can accept a write it will not
return. Any failure leaves the legacy slots untouched and asks the user to retry.

### #950 — keystore residue
(a) `dev-keystore.json` → `keystore.pks`, as write-new → read-back → unlink-old, so
an interruption leaves both files rather than neither. (b) A host that later gains
a Secret Service now **stays** on the file while the file holds keys for the
current namespace, instead of flipping to an empty keychain and reading as
signed-out; namespace matching is exact across all four namespace shapes.
(c) A store that decrypts but does not parse is left exactly where it is — the old
`.bad-{ts}` rename looked like a backup and behaved like a wipe, because the next
operation found no store and the first write established an empty one.

### #945 — the pinned ML-DSA key
`transparency.rs` is now the single source of truth; `scripts/check-pinned-log-key.py`
asserts the other copies agree and **fails on an unregistered new copy**, so the
count cannot silently become ten. Gated on every PR via `scripts-check.yml`
(offline), and `website-verify.yml` keeps only the half it alone can do — is the
deployed site serving what `main` says. Its own test suite
(`scripts/test-pinned-log-key.sh`, 13 cases) breaks one copy at a time and asserts
the checker names it; it also pins that the source-of-truth lookup is by const
name, since the previous grep took the first 2624-hex literal in the file and was
one const reorder away from asserting the root key instead.

### #946 — the whitepaper
Reconciled against the code, using `/subprocessors` as the reference. The client
holds no Resend, R2 or LiveKit credential; §4, §8, §8.1, §9.3, §10.1, §11.1, §12
and §13 said otherwise while §1 already said the opposite — the document
contradicted itself. Also corrected: OTP generation/storage/verification is
server-side and salted, comparison **is** constant-time, rate limiting **does**
exist (per-email and per-IP), the baked Turso token is read-only and superseded at
runtime, and two dead source citations. The same stale claim was live on
`website/security.html` and is fixed there too.

### #942 — drift from #917
The client preflight and the DS guard now share one definition of the membership
SQL (`pollis_schema::authz`), with a parity test over a decision matrix asserting
the two reach the same verdict. `std::mem::forget` is gone from `pollis-delivery`
entirely — a `common::TempDb` guard owns the directory, with a test proving it is
actually removed on drop. The Playwright mock no longer lets a roster that omits
the session user read as a non-member. Plus the contributing note about
`pnpm install` in a fresh worktree.

### Verification
All four suites plus schema, green, with per-test revert-verification for #951 and
#950 (each test breaks only when its own guard is removed). Raw numbers in the
merge comment.

### Deployment
`pollis-delivery` changes are a pure refactor — the executed SQL is byte-identical
— but the binary changes, so the DS is redeployed and its live `/version` SHA
confirmed after merge.


actuallydan:
### Raw before/after

Baseline measured on `origin/main` at `d7e869cc` in an isolated worktree; main's
later commit `d7e3eebc` is mobile TypeScript only, so it cannot move these.

| suite | main | branch | Δ |
|---|---|---|---|
| `pollis-core` headless | 621 | 632 | +11 |
| `pollis-core` default | 648 | 659 | +11 |
| `pollis-tui` | 30 | 30 | 0 |
| `pollis-delivery` | 320 | 324 | +4 |
| `pollis-schema` | 3 | 3 | 0 |

0 failed everywhere. The +11 is 3 (#951) + 8 (#950), counted twice because
`pollis-core` runs in both feature configs; the +4 is #942 (2 fixture-guard, 2
parity). Clippy clean at `-D warnings` on all five crates. Frontend `tsc` clean,
44/44 unit tests.

### Revert-verification

Each new guard was removed one at a time and the suite re-run; every case failed
**only** the test that owns it.

| reverted | test that failed |
|---|---|
| #951 read-back verification removed | `an_interrupted_migration_loses_nothing` |
| #951 legacy slots kept | `legacy_slots_are_wrapped_then_removed` |
| #951 delete not scoped per user | `a_user_with_no_legacy_slot_is_unaffected` |
| #950 rename disabled | `the_store_is_renamed_without_losing_a_key` (+2) |
| #950 unlink before verify | `a_rename_that_cannot_complete_keeps_the_old_file` |
| #950 `.bad-{ts}` rename restored | `a_corrupt_store_is_never_moved_aside_or_started_over` |
| #950 namespace matched by bare prefix | `namespace_matching_is_exact_in_every_shape` |

`scripts/test-pinned-log-key.sh` is the same idea for #945 and runs in CI: 13
cases, each breaking one copy of the pinned key and asserting the checker names
that file — including a rotation applied only to `pollis-core` (which must fail
every other surface at once) and an unregistered tenth copy.

One test caught a mistake of mine while being written: the #942 matrix's
empty-requester row asserted `false`, and the parity half passed while the
"is the agreed answer the intended one" half failed — an empty requester really
is a non-member of a real group, and rejecting an unauthenticated caller is
`resolve_actor`'s job upstream, not this predicate's.

### Note on the environment, not the code

An earlier `ui_e2e` timeout was environmental: this box has no swap and 36G of
its RAM in tmpfs, and 32-way cargo parallelism OOM-killed builds and truncated
tmpfs writes. With bounded parallelism `ui_e2e` is 10/10.

Issue #140, opened 2026-04-11
Voice: allow multiple devices from the same user in one room


actuallydan:
## Summary

Today, two devices belonging to the same user cannot coexist in a voice room. LiveKit identities are derived from `user_id` only, so when a second device joins, the server kicks the first one on the identity collision. Change the voice flow to use a per-device LiveKit identity so both devices can participate, and disambiguate duplicate display names in the UI (`usernameA`, `usernameA (1)`).

## Current behavior (latent bug)

`src-tauri/src/commands/voice.rs` `join_voice_channel` builds the LiveKit identity as:

```rust
let token = make_token(
    &state.config,
    &channel_id,
    &format!("voice-{user_id}"),
    &display_name,
)?;
```

When device B joins with the same identity:

1. LiveKit's default duplicate-identity policy disconnects device A and accepts device B.
2. Device A receives `RoomEvent::Disconnected` which becomes `VoiceEvent::Disconnected`.
3. `frontend/src/hooks/useVoiceChannel.ts` handles the event and clears local UI state, but there is no toast or explanation — voice silently ends from the user's perspective.
4. Device A's Rust-side state (`VoiceState.room`, cpal mic stream) is NOT torn down by the event loop — the `Disconnected` branch doesn't wire to `leave_voice_channel`. Resource leak until the user manually leaves or quits the app.

A sibling flow in `src-tauri/src/commands/livekit.rs:256-258` already does this correctly for inbox rooms, using `state.device_id.lock().await.clone()` to build a per-device identity. That comment even says: *\"Build a per-device identity so multiple devices for the same user can…\"*. The voice flow missed the pattern.

## Desired behavior

Both devices join the same voice room as separate participants. In the UI, duplicate display names get a disambiguator suffix (`usernameA`, `usernameA (1)`, `usernameA (2)`). Audio from the current device's *own* sibling devices must be filtered out to prevent self-hear / echo.

## Proposed design

1. **Per-device LiveKit identity.** Change `voice.rs` to build the identity as `format!(\"voice-{user_id}:{device_id}\")`, matching the inbox pattern in `livekit.rs`. Requires `device_id` to be available in the voice command — pass it from the frontend or pull from `state.device_id` like the inbox flow.
2. **Parse identity on receive.** Anywhere the frontend or backend treats `identity` as a user key, split on `:` and take the prefix as `user_id`, suffix as `device_id`. Affects `useVoiceChannel.ts` participant map and `VoiceEvent` emission sites.
[...]

PR #979, opened 2026-08-17
feat(livekit): pseudonymous per-room participant identities (#836)

### LINKED ISSUE #836: LiveKit participant identity still carries raw user_id:device_id
Follow-up to #828, which pseudonymised room names. This is the other half.

## The residual

Room names no longer tell the SFU operator *which conversation* a room is. But participant identity is still the raw `{user_id}:{device_id}` (built server-side in `broker::livekit_token` from the verified signer, with `voice-` / `:view` variants per `kind`).

So the operator still sees **which users share a room**, and can cluster the social graph by co-membership — it just cannot name the conversations. #828 removed the join from room → Turso row; it did not remove the user identifiers themselves.

Same reasoning as #828 applies: academic while we run the only SFU, not academic with regional SFUs or a third-party one.

## Why it was not done in #828

Identity is load-bearing in ways room names are not. It is matched against, not just routed on:

- **Voice roster** — `list_voice_participants` filters internal identities and maps participants back to users.
- **Speaking indicators / presence** — `emit_room_initial_presence` and the participant-connected path key off identity to attribute audio and presence to a user.
- **Per-device coexistence (#140)** — `voice-{user}:{device}` is what lets a user's second device join as a distinct participant instead of colliding.
- The `:view` and `voice-` prefixes encode capability, and the client parses identity in places.

So unlike room names — which the client never reads, because `Room::connect` takes no room argument and the room lives in the JWT grant — identity **is** read by both sides. A pseudonym here needs a reverse mapping available wherever a participant is attributed to a user, which is a real design question, not a substitution.

## Shape of the work

- Decide where the mapping lives. The DS can mint identity pseudonyms, but the *client* needs to resolve them to display a roster — so either the DS returns a per-room identity↔user map to authorized members, or identity stays derivable by members only.
- Preserve per-device distinctness (#140) and the `voice-`/`:view` capability split.
- Keep the invariant that a client cannot mint a token as another user — identity is derived from the verified signer today, and that must survive.

## Acceptance

- LiveKit never sees a raw `user_id` or `device_id` as a participant identity.
- Voice roster, speaking indicators, and presence still attribute correctly.
- A user's two devices still coexist in one room as distinct participants.
- A client still cannot obtain a token identifying as another user — with a test.

Related: #828 (room names, shipped), #489 (per-conversation membership pseudonyms, deferred).

actuallydan:
Closes #836.

#828/#835 made room names opaque. Participant identity was still the raw `{user_id}:{device_id}` (`voice-` / `:view` per kind) and the JWT `name` claim was the user's **username**, so the SFU kept the membership of every room and could cluster the social graph by co-membership — a second, independent copy of it outside Turso — without ever being able to name a conversation. Verified still true on `main` before starting: `broker.rs` built `format!("{user_id}:{device_id}")` and passed `display_name` into the `name` claim.

### The construction

Every identity is now `{prefix}{base64url(siv ‖ ct ‖ tag)}` — the user id **encrypted** under a key derived per **logical room** (`pollis-delivery/src/participant_id.rs`), deterministic AEAD in the SIV shape built from HMAC-SHA256 only (no new dependency; same primitive keying `room_id.rs`). The device id is folded into the synthetic IV, so a user's two devices stay distinct participants (#140) without either id reaching the wire. `name` is empty.

Only a 2-character capability prefix stays legible (`d-` realtime, `v-` voice, `w-` view) — both ends route on it, and capability was already visible before this change.

**Encryption, not a one-way MAC** (unlike room names) because identity is *matched against*, not merely routed on: roster, speaking indicators, presence and the self-hear filter all need to get from a participant back to a user. A MAC forces the resolver to enumerate candidate `(user, device)` pairs — impossible for `call-<ulid>` rooms (no membership rows) and stale for a device enrolled since the last sync.

**The key stays server-side**, exactly as in #828 and for the same reason: clients resolve through the new `POST /v1/livekit/identities`, authorized identically to minting a token for that room, so it discloses nothing the caller could not learn by joining. A per-room key shipped in a release binary could be extracted; one that resolved every room would hand the graph straight back.

Identity is still a pure function of the room, the **verified signer** and its signature-verified device — nothing a client sends moves it — so "a client cannot obtain a token identifying as another user" survives, with a test.

### Client side

`pollis-core/src/commands/livekit_identity.rs` (outside the media gate, so it unit-tests on every target) owns the two identity spaces: the opaque **wire** pseudonym and the internal `voice-{user}:{device}` key the app and renderer already speak. Every LiveKit boundary translates inbound, so **the renderer is unchanged** (docs only). Resolutions are memoized per `(room, wire identity)`.

The local participant's pseudonym is **pinned** at join, read from the `sub` claim of the token we were just handed — resolving our own would return a synthetic-device identity and silently detach the local tile from its own push-to-talk mute events.

Degrades safely: an unresolvable identity is used as-is, so a peer shows as an unattributed tile rather than not showing up.

### Payload paths that would have undone it

`typing` and `voice_joined`/`voice_left` are **shared-room** broadcasts that carried the actor's `user_id`/`username` in cleartext — published by the very participant being pseudonymised, several times a minute while composing. That is the shortest possible path back to the account, and it also violated the §5 routing-only rule those rooms are already under. They now carry no identity; recipients attribute them from the **publishing participant** (which is also strictly stronger — a self-declared `user_id` in the body was unauthenticated). Builders moved into `livekit_signalling.rs` so the existing `assert_no_identity` sweep covers them.

`device_revoked` loses its `device_id`/`user_id` too: **no receiver ever read either** (the handler states the payload is advisory and spoofable, ignores the device id, and re-checks authoritatively), and its `user_id` re-identified the inbox room #828 had just made pseudonymous.

### What the SFU can and cannot correlate now

**Cannot:** map a participant to a Pollis account, username or device; tell that a participant in room A is the same person as one in room B (the co-membership clustering #836 was filed for); tell that two rooms share members.

**Can, still:** count participants in a room and see traffic patterns; recognise a returning participant **within one room** — these are stable pseudonyms, not unlinkable ones, exactly as room names are. For 1:1 calls that is already per-call, since `call-<ulid>` rooms are ephemeral; for a persistent voice channel it is not.

Per-call rotation for channel rooms is **not** taken here, deliberately. The only epoch the DS could observe without inventing state is the LiveKit room's creation time, which costs a `ListRooms` round trip on the join hot path and breaks duplicate-identity eviction: a mid-call reconnect landing on a new epoch leaves the stale participant to time out beside the new one — a duplicate tile where there is none today. Same call #828 made for room names.

**Out of scope, stated rather than implied:** the private-inbox payloads (`call_invite`'s `caller_id`/`caller_username`, `dm_created`, group invites) still name people. That is an explicit, documented product decision in `metadata-minimization-design.md` §5 — inbox payloads are deliberately carved out of the routing-only rule so alerts can say who is calling — not an oversight of this ticket, and reversing it for one of the three would be incoherent.

### Version skew

New client + old DS: fully compatible — `/v1/livekit/identities` 404s, resolution falls back to the identity as-is, which *is* the old-style internal identity.

Old client + new DS: degrades cosmetically. Live voice tiles lose names/avatars (the SDK identity is opaque to it and `name` is empty) and typing indicators stop, until the client updates. The roster endpoint still returns resolved names, and nothing functional breaks — calls connect, audio flows, messages deliver. Called out because a clean cut is the only option that actually satisfies "LiveKit never sees a raw user_id"; a capability flag would leave the leak open with no forcing function.

### Tests

`participant_id.rs` — 9 unit tests: never embeds the user/device/room for any kind, round-trips to user+kind, stable within a room, **unlinkable across rooms** (and room A's key does not open room B's identity), two devices distinct but same user, kinds are independent strings, distinct users do not collide (incl. the length-prefix concatenation guard), re-keys with the secret, rejects forged/malformed/legacy identities, empty device supported.

`tests/broker.rs` — 4 more asserted on the **decoded JWT** (the thing LiveKit reads, so a pseudonym applied anywhere short of it still fails): `sub` embeds no user/device/room for any kind and the whole token contains neither; one user unlinkable across two rooms; two devices coexist and resolve to one user; identity determined by the signer, not the request. Plus `name` is now pinned empty.

`livekit_identity.rs` — 11 unit tests on the translation: round-trips under the same helpers the app uses, two pseudonyms of one user stay distinct participants, translation is stable per pseudonym, `:view` survives (and an *unresolved* `w-` is still filtered), unresolved pseudonyms are not mistaken for users, internal participants are not users, local identity shape, the self-hear predicate, degenerate pseudonyms still produce a key, and `jwt_subject` for the local pin.

`livekit_signalling.rs` — 3 more: typing and voice-presence broadcasts carry no actor, `device_revoked` carries only its type.

Raw numbers on `origin/main` d7e869cc → this branch:
- `cargo test -p pollis-core --no-default-features --features test-harness`: **621 → 635**, 0 failed
- `cargo test -p pollis-delivery`: **320 → 334**, 0 failed
- `cargo test --features test-harness --test flows`: **95 passed, 0 failed**, 1 ignored
- `cargo clippy --workspace --all-targets -- -D warnings`: clean

**Coverage honesty:** the browser tier has no LiveKit, so "local participant is still recognised as local", "peers attributed to the right person", and per-device coexistence are covered at the **Rust** level (identity translation, self-hear predicate, local-pin) plus **inspection** of the renderer's unchanged parsing — not end-to-end against a real SFU. Renderer changes are comment-only.


actuallydan:
### e2e verification note

Both two-client LiveKit e2e runs on the first push failed — investigated, and the cause was **pre-existing infrastructure, not this change**:

```
[apply-log-migrations] applied 0 statement(s) from migrations-log/
[ingest] process_pending_commits: SQL input error: no such column: generation
[mls] reconcile: remote persist failed: delivery submit 500 Internal Server Error
[two-client-call] FAILED: A: the other call participant never appeared
[voice-channel] FAILED: timed out waiting for [data-testid="voice-channel-view"]
```

`e2e/scripts/apply-log-migrations.py` globbed `pollis-core/src/db/migrations-log`, but those migrations moved to `pollis-schema/migrations-log/` in #920. The glob matched nothing, the script printed `applied 0 statement(s)` and **exited 0**, so the commit-log tables silently stayed at baseline — reintroducing precisely the breakage its own docstring says it exists to prevent ("cross-client MLS delivery silently breaks"). Group/DM setup then 500s and no second participant can ever appear, which is why both voice e2e tests failed downstream.

Fixed in the second commit: correct path, plus a guard that makes an empty glob a hard error instead of a silent success. It now finds all 5 files / 22 statements. A baseline run of the same workflow on unmodified `main` is running to confirm the failure predates this branch.

actuallydan:
### Update: second defect in the same script, and the baseline proof

**Baseline confirmed.** The same workflow on unmodified `main` (run 32083152823) fails identically:

```
[apply-log-migrations] applied 0 statement(s) from migrations-log/
[ingest] process_pending_commits: SQL input error: no such column: generation
[two-client-call] FAILED: A: the other call participant never appeared
```

So the e2e failure predates this branch and is not caused by the identity change.

**Second defect, surfaced by fixing the first.** With the path corrected, the applier failed loudly:

```
SQL string could not be parsed: unexpected end of input at (11, 90)
SQLite error: cannot commit - no transaction is active
```

`statements()` split on `;`, but `000005_mls_commit_log_triggers.sql` defines three `CREATE TRIGGER ... BEGIN <stmt>; <stmt>; END;` bodies whose internal semicolons shred into fragments — and the orphaned `END` parses as `COMMIT`, hence the second error. The splitter is now trigger-aware.

**Checked the deploy path:** `scripts/db-apply.sh` — the one that applies real migrations to dev/prod — already re-joins `CREATE TRIGGER` fragments up to `END`. **Production was never affected**; only the e2e fixture had the naive split.

**Regression guards** (a case that fails on the old splitter):
- `--self-test` with an embedded trigger whose body carries its own semicolons: naive split → 5 fragments, 1 incomplete; correct split → 3 statements, 0 incomplete.
- Every emitted statement is validated with `sqlite3.complete_statement` before anything is sent, so a bad split fails locally and by name instead of as a server-side `unexpected end of input`.
- An empty glob is now a hard error rather than `applied 0` + exit 0.
- Both wired into `scripts-check.yml`, which also asserts the real files still split into complete statements (5 files / 19 statements).

actuallydan:
### Real end-to-end verification: both LiveKit e2e runs pass

With the fixture repaired, both two-client workflows pass **on this branch**, against a **real LiveKit SFU** and a DS running with **`POLLIS_DS_REQUIRE_AUTH=true`** — i.e. identities derived from real device signatures and resolved through `/v1/livekit/identities`.

[Voice channel](https://github.com/actuallydan/pollis/actions/runs/32084035941) — success:
```
[apply-log-migrations] applied 19 statement(s) from migrations-log/
[voice-channel] A: sees 2 participant(s)
[voice-channel] B: sees 2 participant(s)
[voice-channel] A: left (join CTA reappeared)
[voice-channel] B: sees 1 participant(s)
```

[1:1 call](https://github.com/actuallydan/pollis/actions/runs/32084037704) — success:
```
[two-client-call] A: sees 2 participants in the call
[two-client-call] B: sees 2 participants in the call
[two-client-call] SUCCESS: both clients see each other in the call
```

This upgrades the coverage claim in the PR body. Now demonstrated end-to-end rather than by inspection:

- **The local participant is recognised as local** — each client renders its own tile and the join/leave CTAs track it, which is the `pin_local_identity` path.
- **Peers are attributed to the right person** — each side resolves the other's opaque pseudonym to a user and renders a second tile; a wrong or unresolved identity would show as an unattributed/duplicate tile or none at all.
- **Leave is attributed correctly** — B drops to 1 participant when A leaves, so `ParticipantDisconnected` translated to the same internal identity the join produced.
- **The `call-<ulid>` path works** — the room shape with no membership rows, which is exactly why the pseudonym is reversible rather than a MAC.

Still only covered at the Rust/unit level (no LiveKit in the browser tier): two devices of one user coexisting in one room (#140 — covered by unit tests on the pseudonym and the translation, not by a live second device), and cross-room unlinkability (a property of the construction, asserted in `participant_id.rs` and on the decoded JWT).

actuallydan:
### Rebased onto `0e350240` — baselines corrected and gates re-run

Rebased past #977 and the mobile merge train onto current `origin/main` (`0e350240`). One conflict, in `scripts-check.yml`: #945 added its pinned-transparency-key steps where this PR adds the migration-splitter step. Purely additive — both kept, and `check-workflows.py` re-validated.

**Corrected raw numbers.** The earlier figures were measured against the pre-#977 base, and my first re-baseline attempt was contaminated (two worktrees sharing one `CARGO_TARGET_DIR` clobber each other's path-dependency artifacts). Re-measured properly, both baseline and branch in the **same worktree path and target dir**:

| gate | `origin/main` @ `0e350240` | this branch | delta |
|---|---|---|---|
| `cargo test -p pollis-core --no-default-features --features test-harness` | 632, 0 failed | **646**, 0 failed | +14 |
| `cargo test -p pollis-delivery` | 324, 0 failed | **338**, 0 failed | +14 |
| `cargo clippy --workspace --all-targets -- -D warnings` | clean | **clean** | — |

DS-affecting gates re-run on the rebased tree because this change touches token minting and the identity resolver: the full `pollis-delivery` suite, the pollis-core suite, workspace clippy, and the migration-splitter self-test all pass. Both two-client LiveKit e2e workflows have been re-dispatched against the rebased tree.

**Deploy target corrected:** both environments are currently live at `0e350240`, not `ba7e9ea7`. Post-merge verification will be against the new `main` SHA this PR produces.

PR #984, opened 2026-08-19
Custom emoji upload: drag-and-drop target, filename-derived shortcode, live preview

actuallydan:
Reworks the Custom Emoji page to Discord's model — image first, name derived — and fixes the layout.

- **Drop target.** New `EmojiDropZone` (`components/Emoji/EmojiDropZone.tsx`): drop a gif/image anywhere on the window, or click to open the picker. Registers as an *inline* drop target so `AppShell`'s full-window overlay stays out of the way and the zone lights up itself (`AppShell` now rebroadcasts drag state as `pollis:pathdragstate`; `dropTargetStore` gained a second, overlay-free counter). A native drag never reaches the DOM under Tauri, so "drag-over" is a window-level fact, not a hover.
- **Order is no longer forced.** The shortcode is auto-filled from the filename, sanitised to `[a-z0-9_]` (lowercased, extension stripped, illegal runs collapsed to `_`, trimmed, capped at 32). A field the user has already typed in is never clobbered, so typing the name first still works.
- **Preview.** A blob URL of the source file, shown at 4rem in the zone and again at the 1.375rem it renders at inline, next to a live `:shortcode:` sample. Gifs animate in the preview.
- **Alignment.** The add button no longer sits in a row with an input that has a helper line under it — it has its own row.
- **Inline validation.** Empty / invalid / duplicate shortcode go through `TextInput`'s error slot; unsupported file type and over-8 MB source go to the page error line. Add is disabled until valid.

Upload semantics are untouched: same `upload_group_emoji` call with the same `{groupId, shortcode, path}`, same 48 KB re-encode and content-hash dedup in Rust, and animated gifs are already re-encoded as gifs (`encode_emoji`'s frame ladder), so animation is preserved as before. The re-encoded size is not knowable before the upload, so the pre-upload figure is labelled as the SOURCE size.

Docs: `.codesight/wiki/ui.md` regenerated, with notes on `EmojiDropZone` and `GroupEmoji`. New unit test `frontend/tests/emoji-shortcode.test.ts` pins the filename→shortcode sanitiser against `SHORTCODE_RE`. New copy translated in all seven locales.

`pnpm exec tsc --noEmit` clean; `pnpm test` 52/52.

CI note: `supply-chain` is red on RUSTSEC-2026-0258 (`h2 0.3.27`, transitively via `libsql` → `tonic` → `hyper 0.14`). Newly published advisory, unrelated to this frontend-only diff — it needs its own dependency ticket.


actuallydan:
Post-merge release checklist (blast radius: `frontend/` only — no Rust crate, no schema, no wire contract):

- Desktop app — **defer**: user-visible, ships with the next `v*` tag; nothing here needs an out-of-band release.
- CLI (`pollis-tui`) — N/A
- DS (`pollis-delivery`) — N/A
- Relay pool / `pollis-verify` / transparency tree — N/A
- DB migrations — N/A (none in this PR)
- Website / LiveKit — N/A

Separately: `supply-chain` went red on this PR for RUSTSEC-2026-0258 (`h2 0.3.27`, transitive via `libsql` → `tonic` → `hyper 0.14`), a newly published advisory unrelated to this diff. It needs its own dependency ticket — `cargo update -p h2` cannot cross the 0.3 → 0.4 line here.

---

PR #985, opened 2026-08-19
Breadcrumb back/forward chevrons driven by router history

actuallydan:
The breadcrumb bar's single back arrow navigated *up the breadcrumb tree* and vanished at the app root. It is now a browser-style back/forward chevron pair over **router history**, the way Slack and Discord do it — one entry per click, no hierarchical navigation left in the arrows (the trail's own segments still jump up the tree).

Both chevrons are always rendered, root included, and each is a real `disabled` `ui/Button` (`variant="ghost"`) when that direction has nowhere to go — so the grey and the accessibility tree agree, and nothing shifts sideways as the trail shortens.

**Knowing "can I go forward?"** — no browser API answers it: `history.length` counts a session, not a direction, and never shrinks when a push discards the entries ahead. So `utils/historyNav.ts` keeps a cursor and a ceiling off the history's own `__TSR_index`: `canGoBack` is `index > 0`, `canGoForward` is `index < maxIndex`, and a PUSH *resets* the ceiling to the new entry because that is exactly what a push does to the forward stack. Pure, and pinned by `frontend/tests/history-nav.test.ts` (10 cases: clamping at both ends, push truncation vs. replace, multi-entry GO, unstamped/nonsense indices). `stores/historyNavStore.ts` is the MobX shell — subscribed to `router.history` (event-driven, no polling), attached once in `TerminalApp` next to the router rather than in `BreadcrumbNav`, which unmounts on skin changes and would forget the forward stack each time.

**`nav.back` (Escape):** kept, and pointed at the same `historyNavStore.goBack()`. Its channel special case — from a channel, jump to the group page "to avoid landing on create-channel if that was in history" — is gone: that was hierarchical navigation wearing a history label, and a shortcut that lands somewhere other than the button beside it is worse than walking back through an entry a browser would walk through too. It also now no-ops at the first entry instead of calling `history.back()` into nothing. The screen-share-viewer exit still takes precedence.

`e2e/emoji-picker.js` climbed channel → group via that back arrow; it now clicks the trail's group segment instead. `e2e/two-client-channel.js`'s back click (invite page → group page) is unaffected — there the previous history entry *is* the group page.

Also: `common:actions.forward` added to all seven catalogues; `.codesight/wiki/ui.md`'s `BreadcrumbNav` entry rewritten.

Verified: `tsc --noEmit` clean, `pnpm --filter frontend test` 62/62, `scripts/i18n-check.mjs` OK, `scripts/ui-inventory.mjs --check` OK.

CI note: `supply-chain` is red on `main` itself (a newly-published RUSTSEC advisory against unchanged transitive deps — `advisories FAILED, bans ok, licenses ok, sources ok`), not caused by this PR, which touches no Rust and no `Cargo.lock`.


actuallydan:
Second commit: `e2e/emoji-picker.js` climbed from a channel to its group by clicking the breadcrumb back arrow. With the arrow now walking history, the entry behind that channel is the create-channel form it was made from, so the script clicks the trail's group segment instead — hierarchical jumps live in the trail now. `e2e/two-client-channel.js`'s back click (invite page → group page) is unaffected: there the previous history entry *is* the group page.

CI note: `supply-chain` is red on `main` itself (a newly-published RUSTSEC advisory against unchanged, transitive deps — `advisories FAILED, bans ok, licenses ok, sources ok`), not caused by this PR, which touches no Rust and no `Cargo.lock`. Merge-blocking `gate` and the renderer `frontend-check` are green.

---

actuallydan:
Post-merge release checklist (blast radius: `frontend/` only — no Rust, no `Cargo.lock`, no `pollis-schema` migration, no DS code):

- Desktop app — **defer**: user-visible, but it rides the next `v*` tag / `desktop-release.yml` like any other renderer change; nothing here needs an out-of-band release.
- CLI (`pollis-tui`) — **N/A** (no `pollis-core` change).
- DS (`pollis-delivery`) — **N/A** (no wire contract, no DS code).
- Relay / verifier / transparency log — **N/A** (`verifiable-log*`, `pollis-relay`, `pollis-device-cert` untouched).
- DB migrations — **N/A** (none added).
- Website / LiveKit — **N/A**.

Merged with `supply-chain` red; that check is red on `main` itself (RUSTSEC advisory against unchanged transitive deps) and is tracked separately. `gate` and `frontend-check` were green on the merged head.

---

PR #988, opened 2026-08-20
Move every client-side remote read behind the Delivery Service, then delete the client's database credentials (#987)

### LINKED ISSUE #987: Move all client remote reads behind the DS, then delete the client's database credentials
Move every client-side remote read behind the Delivery Service, then delete the client's database credentials entirely.

Today the desktop/CLI/mobile client reads Turso **directly** — 102 `remote_db` call sites across 34 files, plus 12 `log_db` sites in 7 — using a whole-DB read-only token baked into the binary. Writes already go through the DS. This closes the other half.

---

## 1. Why — and why now

This came out of an architecture review of four goals: the DS's eventual single-writer bottleneck, dropping Durable Objects (cold start + it contradicts the run-it-yourself ethos), avoiding vendor-specific backend logic, and cutting round-trip latency.

Three of those are present-tense; the single-writer one is explicitly "eventually". Every candidate answer to the *eventual* one — Postgres, Cassandra/Scylla, distributed SQL, per-conversation sharding — is a one-way door bet on a workload we cannot observe yet, with no users. That is the same shape that produced #261 and got it closed not-planned after a spike, an fd prototype and a `ReplicaPool` skeleton.

**So this ticket deliberately does not pick a database.** It is the move that makes that decision cheap, reversible and *later*, while paying for itself immediately on three independent axes:

**Latency, today.** Reads are sequential and *dependent* — each query's input is the previous query's output, so the client cannot pipeline them. Measured from the audits:

| Path | Dependent round trips today |
|---|---|
| Channel open / realtime nudge | 5–7 |
| DM open | 6–7 |
| Send (group) | 8–10+ |
| Send (N-party DM) | 8–10 + **one per peer**, sequential |
| Cmd+K palette open | 1 + 2N (a gate query *and* a member query per group) |
| Invite one user | 6 |
| **Cold launch (8 groups + 12 DMs)** | **~60–80, serially** |

Co-locating data reduces each hop's cost linearly. Running the chain server-side next to the DB reduces the *count*. The second is the bigger number, and it needs no sharding, no replicas and no vendor features.

**Supply chain — concrete, not hypothetical.** `supply-chain` CI is red on `main` right now from a RUSTSEC advisory reaching the **shipped client binary** transitively through libsql → tonic → hyper 0.14. Removing client reads is what lets libsql leave the client dependency graph. See §6 for the two edges that must both be cut for this to actually happen.

**Read authorization.** The client holds a whole-DB read token: any user can `SELECT` every membership row and every envelope's ciphertext and timing. #393 closed the *write*-credential leak and explicitly left this open; #261 identified it as the strongest argument for per-conversation DBs and then **forfeited it** on close ("Accepting the whole-DB read token for now"). This closes it structurally and more completely than #420 did — a client with no credential at all is strictly stronger than one with a read-only token.

**And the deferral property.** Once the database sits entirely behind the DS, swapping engines is a server-side change with **no client involvement** — no fleet release, no version straddle. Compare the ordering hazards we have had to document for #607, #668 and #669. Deferring a decision usually makes it more expensive; this is the rare case where the deferring move makes the deferred decision *cheaper*.

**Related:** #261 (per-conversation DBs, closed not-planned) · #420 / `docs/goal-a-commit-log-sole-writer.md` (commit-log sole writer) · #393 (secrets broker, read-scope residual) · #532 / #540 (directory index, added then reverted) · #539 (commit-log pruning) · #658 (DS latency spike) · #484 / #877 (reproducible builds)

---

## 2. What the audit found — including three corrections to current docs

Five parallel read-only audits covered messages/ingest, directory/groups/DMs, the MLS control plane, auth/devices/bootstrap, and the credential surface.

**Correction 1 — the #532 directory index was reverted and its tables are stale.** PR #540 removed `pollis-delivery/src/directory.rs` and every maintenance call site, and flipped `list_user_groups_with_channels` back to the `group_member` JOIN. The DS now only ever `DELETE`s from `user_groups` / `user_dms` — never inserts, updates or selects. They hold the one-time migration backfill minus deletions. **Any endpoint reading them would silently serve a wrong sidebar.** Derive from `group_member` / `dm_channel_member`.

**Correction 2 — commits and envelopes are already in different databases.** Since #420, envelopes live on the main DB (`remote_db`) and commits on the log DB (`log_db`) — two independent Turso databases with independent replication. The commit-vs-envelope pair is therefore **already eventually consistent**, and `catch_up_mls_group_interleaved` already tolerates it. Moving commits behind HTTP loses no guarantee here.

**Correction 3 — the #420 read-site table (R1–R6) is stale.** Every line number moved, R1 was split into a separate function, the status board's "R6 deferred" is wrong (it resolved when W5 became `POST /v1/welcomes/ack`), and **three read sites are missing entirely**: `published_group_info_head`, `has_pending_welcome`, and `migrate.rs`'s `head_epoch_in`. Actual log-DB count is 8, not 6.

**Scope is smaller than "move all reads".** The message render path is *already local* — pagination, search, threads and receipts all read decrypted plaintext from SQLCipher. There is no "fetch messages" endpoint to build. The credential is held for **ingest, send/edit/delete preflight, and the directory**.

**`GET /v1/commits/:conversation_id` has zero client callers.** It is a shape to imitate, not a live path to migrate.

---

## 3. Hard design constraints

These are not preferences. Getting any of them wrong breaks message delivery or security.

### 3.1 GroupInfo and pending-Welcome MUST share one snapshot

`external_join_attempt` reads GroupInfo, then checks for a pending Welcome eight lines later. The correctness argument is written into the source: the DS writes commit, GroupInfo and Welcomes in **one IMMEDIATE transaction**, so a reader that observes the GroupInfo must also observe the Welcome. That holds across two reads of one libsql database. It does **not** hold across two HTTP requests that may hit different replicas or different moments.

Break it and the device external-joins while its own Welcome is inbound. Both paths do "delete local group → rejoin", they clobber each other, and the member is **permanently stranded** on `no GroupInfo stored — cannot external-join`. The Kani property `ExternalJoin ⟹ !welcome_pending` is proved over a *pure function* and becomes **vacuous** if the runtime feeds it inputs from two snapshots — machine-checked correctness that silently stops applying.

The same pairing recurs at `process_pending_commits_locked_impl` (the two inputs to the `join_recovery` gate, evaluated eight lines apart) and at migration adoption.

**Requirement: one request, one read transaction, Welcome check evaluated at or after the GroupInfo read.**

### 3.2 The commit batch MUST be contiguous

A non-contiguous batch is treated as a permanent gap, and the response is **destructive** — `forget_local_mls_group_at` drops the device's MLS crypto state. The in-source justification is literally *"the commit log is append-only and Turso reads are consistent"* — a claim about the transport, which is exactly what an HTTP boundary invalidates.

- The `commits` array must be hole-free in `epoch` from first to last element within one `(conversation_id, generation)`.
- If it ever paginates, page boundaries must be exact. A client that receives `[E … E+k]` then `[E+k+2 …]` destroys its group.
- `head`, `head_generation` and `commits` must come from **one read transaction**. **Today's `GET /v1/commits` handler does not do this** — three separate queries on a bare connection. `submit_commit`'s rejection path does it correctly, inside the IMMEDIATE transaction. Copy the second. Extending the first propagates a torn-read bug onto hot paths.
- Add `pruned_below` so a legitimate retention prune (#539) is distinguishable from a torn read.

### 3.3 Read endpoints MUST be POST with signed bodies

The canonical signing message is `{METHOD}\n{PATH}\n{TIMESTAMP}\n{HEX_SHA256_BODY}`, where PATH is **"the request path only, no query"**. A signed `GET …?since=N` therefore carries unauthenticated parameters — precisely the shape of #681, where an unauthenticated `GET /v1/commits?user_id&device_id` was a remote commit-log wipe. Also: GETs are exempt from rate limiting, and `ENDPOINTS` in `pollis-api` excludes them, so a GET read endpoint is invisible to the route-coverage test that exists to catch this class of omission.

Keep `GET /v1/commits/:conversation_id` unchanged for old clients and operator debugging.

### 3.4 Asymmetric failure biases must survive

Each read encodes a deliberate bias, and **HTTP makes failures common where libsql failures were rare**:

| Site | Bias on failure | Break it and… |
|---|---|---|
| `our_commit_is_canonical` | `None` → Rollback | adopting on failure = phantom epoch, undecryptable messages |
| `published_group_info_head` | absent → republish | assuming durable on a blip = group unjoinable forever |
| `has_pending_welcome` | **fail-OPEN** | fail-closed = recovery deadlock |
| `local_user_is_member` | **fail-CLOSED** | fail-open = removed member rejoins and decrypts post-removal traffic |
| roster-device KP query | silently short list | truncated body is indistinguishable from "no KP" |

`has_pending_welcome` and `local_user_is_member` are evaluated ~7 lines apart with **opposite** biases. One endpoint answering both must not collapse them into a single error path.

Related: both `local_device_registered` and `local_user_is_member` justify their fail-closed safety with *"the same `remote_db` was already read successfully for the commit-log fetch, so an error here is vanishingly unlikely."* **That sentence becomes false** once the commit fetch is an HTTP call, turning both into realistic lockout paths. Update the reasoning or re-examine the gates.

### 3.5 Decisions stay client-side

`invariants::resolve` / `classify` / `join_recovery` / `may_rejoin` are Kani-proved pure decision cores. The DS supplies **inputs**, never decisions. Moving a decision server-side silently deletes a machine-checked proof. Likewise the commit-log CAS stays authoritative — no read may replace a server-side re-validation (`migrate.rs`'s `closes_epoch` is the model: read advisory, write re-validates).

---

## 4. Proposed endpoints

All POST, all declaring a `pollis-api` request struct with a typed `Response`, all routed via `<Body>::PATH` so the existing route-coverage test sees them.

| Endpoint | Collapses | Authorization |
|---|---|---|
| `POST /v1/conversations/catch-up` | membership resolve → is_dm → channel enumerate → envelope fetch (+ DM peer keys) | `writes::is_member`; **`device_id` bound to the authenticated device** |
| `POST /v1/mls/conversation-state` | GroupInfo head + blob, pending-Welcome flag, commit batch, head/head_generation, commit-at-epoch | `writes::is_member` + `resolve_recipient` |
| `POST /v1/welcomes/fetch` | the Welcome drain | `resolve_recipient` — recipient must equal authenticated user |
| `POST /v1/directory/bootstrap` | groups+channels tree, DM list, DM requests, pending invites, admin join-requests, blocks, preferences | actor-scoped; admin sub-arrays gated by `is_admin` |
| `POST /v1/directory/group` | one group's members/channels/invite-links/join-requests/emoji | `is_member`; `is_admin` for links + requests |
| `POST /v1/directory/members` | the Cmd+K N-group fan-out | `is_member` per id |
| `POST /v1/directory/users` | profile + search + username hydration + avatars | authenticated; **drop `phone` from the response** |
| `POST /v1/directory/group-by-slug` | slug discovery | deliberately none — but **add a rate limit** |
| `POST /v1/directory/conversations` | id-only enumeration for `auth.rs`, `mls/sweep.rs`, `device_enrollment.rs` | actor-scoped |
| `POST /v1/messages/lookup` | envelope→conversation resolve | `conversation_for_message` + `is_member`; return `None`, not 403, so it is not an existence oracle |
| `POST /v1/read/account-key` | safety numbers, TOFU pinning, cross-signing verification — 6 call sites, batched | device-signed |
| `POST /v1/read/devices` | device list, self-registration check, stale-cert candidates, added-device cert verification | device-signed |
| `POST /v1/read/enrollment-request` | enrollment poll + approve | **capability (`request_id`)**, see §6.5 |
| `POST /v1/read/recovery-blob` | Secret-Key recovery | OTP session only, never plain; rate-limit hard, log a security event per fetch |
| `POST /v1/read/emoji` | usable/group emoji + object lookup | device-signed; server enforces `require_member` |
| `POST /v1/read/voice-room` | 4 reads + 2 sequential chains on voice join | device-signed |

**Every authorization gate these need already exists in `pollis-delivery`** — the DS enforces the identical predicate on the write half of the same tables. No new authz primitive is required.

**The biggest wins are deletions, not additions.** Most client preflights are already re-derived server-side (there is a `preflight_parity` test proving the verdicts match). Fold them into typed rejection reasons on existing write responses and the round trips vanish with no new endpoint:
- `CreateInviteBody` takes an identifier instead of a resolved id → 6 round trips become 1
- `CreateJoinRequestBody` absorbs its 3-read preflight
- `UpdateGroupBody` / `UpdateChannelBody` return the updated row
- `AcceptInviteBody` returns `group_id`; `RedeemInviteLinkResponse` gains `group_name`
- `CreateDmBody` does dedupe + block checks server-side and returns the full DM with members
- `publish_key_packages` computes the shortfall itself → deletes a client read *and* a TOCTOU
- `VerifyOtpResponse` gains `account_id_pub`; `VerifyEmailChangeBody` gains `username`; `RevokeDeviceBody` returns affected conversation ids
- **Push fan-out moves server-side entirely** — `/v1/messages/send` already knows the conversation, sender and mention list. This is the largest hot-read removal available and it takes `push_token` out of client visibility.
- `R2PresignResponse` gains `object_exists` — kills both dedup probes and a round trip on the hit path

**Delete rather than port** (all caller-less; the first two are also semantically dead — they filter `WHERE sender_id = ?`, which since #607 sealed sender is always the literal `"sealed"`, so they can never match): `list_messages_by_sender`, `list_channel_previews`, `list_user_groups`, `list_group_channels`, `get_reactions` (renderer removed in #874), the no-DS legacy email-change branch, and the voice-join commit-log diagnostic dump (which additionally has a silent main-DB fallback that must not be copied).

---

## 5. Batch the sweep

The cold-launch sweep runs the whole catch-up chain **per conversation, serially** — the ~60–80 round trip figure above. `conversation-state` needs a batch form taking a list of `(conversation_id, generation, since)`, or a 20-group launch becomes 20+ sequential signed round trips and we have moved the problem rather than fixed it.

Same warning for added-device cert verification, which runs per add-carrying commit during replay: batch it across a catch-up, or fold the cert columns into the commit-log response.

---

## 6. Credential removal — and the two findings that silently defeat it

End state: **the shipped client holds no database credential at all.**

Names to delete (client): `TURSO_URL`, `TURSO_TOKEN` (fed from `secrets.TURSO_RO_TOKEN`), `LOG_DB_URL`, `LOG_DB_TOKEN` — baked via `option_env!` in `config.rs`, injected by three `Load production secrets` steps in `desktop-release.yml` and three in `cli-release.yml`, mirrored in `rebuild-verify.yml`'s recipe `vars`, and surfaced to mobile through `InitConfig` / `EXPO_PUBLIC_TURSO_*`.

Names that **stay**: DS-side `TURSO_URL` / `TURSO_TOKEN` (read-write), `TURSO_ADMIN_TOKEN`, `LOG_DB_URL`, `LOG_DB_ADMIN_TOKEN` — used by `apply-migrations`, `db-migrate-dev.yml` and the DS deploys. **`TURSO_URL` and `TURSO_TOKEN` are the same names with two different consumers; deleting the DS ones breaks migrations.**

Also dies, last: `POST /v1/turso/token` and its `TURSO_PLATFORM_TOKEN` / `TURSO_ORG` / `TURSO_DB`. These exist *solely* to mint client read tokens.

> **Note:** there is no 401→re-mint loop to delete. The only re-mint trigger is a timer; `is_transient_libsql_error` does not match auth failures, so an expired minted token would fail with no automatic recovery. Removing client reads removes that hazard class.

### The two findings most likely to be missed

**(a) libsql reaches the client through *two* edges.** The direct one is `pollis-core/Cargo.toml:73`. The other is `pollis-core → verifiable-log-serve → verifiable-log-builder`, where libsql is in `[dependencies]` (verified). **Cutting only the direct edge leaves libsql, tonic and hyper 0.14 in the shipped binary and the supply-chain failure unchanged.** `verifiable-log-builder`'s libsql use is the Turso-reading *builder*; the client needs only the verifier half, so this wants a feature gate or a crate split.

**(b) hyper 0.14 is declared *directly* in `pollis-core`.** Lines 131–133 declare `hyper 0.14`, `hyper-rustls 0.25` and `tower-service`, with a comment justifying them as free riders already present via libsql. Their only consumer is `net::overlay::overlay_connector`, whose only caller is `db/remote.rs:74`. Once libsql goes they are the **sole remaining source of hyper 0.14** in the client and must be removed in the same change — otherwise the advisory stays.

What disappears from the client graph when both are cut: `libsql` + its 5 sibling crates, `tonic 0.11`, `hyper-timeout`, `axum 0.6`, `tower 0.4`, `tower-http 0.4`, `http 0.2`, `http-body 0.4`, the `h2` 0.3 line, and `rustls-webpki 0.102` — which is what the three libsql-attributed ignores in `deny.toml:48-50` exist for.

### CI edit order is enforced

`scripts/check-build-recipe.py` asserts `release exports ⊆ rebuild recipe ⊆ config.rs option_env!`. Remove each credential **from the release workflow first, then the rebuilder, then `config.rs` last**. Reverse it and CI goes red between two merged commits. The tripwire parses only the `build-linux` job, so the macOS and Windows blocks are unguarded — grep them by hand.

### `RemoteDb` has no query helpers

It is a connection manager only — every SQL string is written at its call site, and `conn()` has 63 callers. There is no wrapper to intercept, so this is a call-site migration, not a helper rewrite. Note also that `query_only_view` — the harness gate proving the client never writes directly — becomes **vacuous** once the client holds no credential; replace it with a "`pollis-core` declares no libsql dependency" test or it silently stops testing anything.

---

## 6.5 The bootstrap dead end — solve this before writing any endpoint

`ds_post` **cannot be used unless `state.local_db` is `Some`** — it opens a `PollisProvider` over the local SQLCipher DB to load the ML-DSA-44 device signer, and errors *"not signed in for DS request signing"* otherwise. The local DB opens only in `pin::set_pin` / `pin::unlock`, i.e. **only after the user enters their PIN.** So every read that runs before that needs `ds_post_session` (an OTP-session bearer) or `ds_post_plain`.

Most pre-cert reads are fine — the recovery blob, the email confirm, identity-version-for-first-cert and the orphan check all have a session in hand, and `publish_device_cert` already implements the dual gate to copy.

**Two reads have no credential available at all.** `get_session` runs at every app launch, before PIN entry: the local DB is closed, `device_id` is not yet set, and no OTP ran, so there is no session either. It reads user existence, and then `account_id_pub` to compute `enrollment_required` — which gates the pre-unlock UI, so it cannot simply be deferred (a device that needs enrollment has no PIN to unlock with).

**Recommendation:** one `ds_post_plain` endpoint returning `{exists, has_identity}` for a supplied `user_id`, rate-limited by IP. The `user_id` is a 128-bit ULID from the caller's own `accounts.json` — not an enumeration surface — and `verify-otp` already returns `has_identity` for a *guessable* input (an email address), so this discloses strictly less than an endpoint we already ship. Deferring the existence check while keeping the identity check plain is a reasonable hybrid.

### ⚠️ A TTL collision that will brick enrollment if inherited

**Verified:** DS `session_ttl_secs = 600`; client `ENROLLMENT_TTL_SECS = 600`; `await_enrollment_approval` polls for `TTL + 5 = 605s`. The session is minted at `verify-otp`, **strictly before** the enrollment request is created. **A session-gated enrollment poll is therefore guaranteed to 401 before the request it is polling expires.**

Gate that endpoint on **possession of `request_id`** instead (optionally *also* accepting a session). That is cryptographically safe: `wrapped_account_key` is sealed to the requesting device's ephemeral X25519 public key whose private half never leaves process memory, so a `request_id` holder learns nothing usable.

### ⚠️ Two session slots that should be unified first

`ds_post_signed_or_session` reads **only** `bootstrap_session`, which is set only on the first-device path. The soft-reset path runs on a `has_identity` device where only `enrollment_session` was set — so it fails with *"not signed in and no verified-email session"*. Unify the slots (or add `ds_post_session_any`) before routing more reads through that helper.

### A trust separation that must be preserved

Several reads fetch `users.account_id_pub` — the cryptographic root of trust for safety numbers and MLS cross-signing. Today a lying Turso and a lying DS are the same trust domain, so routing through the DS changes nothing *formally*. The mitigation is the transparency log, which already fetches from a **separate origin** (`POLLIS_TRANSPARENCY_URL`). **That separation must be preserved** — if the transparency endpoint is ever collapsed onto the DS, `self_audit_account_key` becomes the DS attesting to itself and the audit is worthless.

---

## 7. Ordering hazard

**The DS ships first, in an earlier cycle than the client, and the client keeps a working fallback until the DS read endpoints are confirmed live in prod.**

**A client that has lost its Turso credential and meets a DS without the read endpoints is not degraded — it is bricked.** Worse than the #607 straddle (which 403'd edits while messages still flowed) and worse than #669 (which mis-bucketed a ciphersuite). Reads are the whole app: sign-in, channel list, history, MLS state. There is no fallback left once `RemoteDb` is gone, because the fail-soft that protects today's design — a mint failure leaves the baked token in place — is precisely what is being deleted.

**There is no data migration.** No column added or dropped, no schema change. Only who executes the SELECT changes.

Order: deploy DS read endpoints (dev → verify → prod, **confirm `/version` reports the merged SHA** — `/health` is not enough) → let them serve real traffic from clients that still hold their baked token and can fall back → cut the desktop tag and dispatch `cli-release.yml` → **credential removal as a third, later step**.

Two asymmetries make this recoverable where #668's was not: the DS side is purely **additive**, so a DS carrying the endpoints is compatible with every client old and new; and an old client meeting a new DS is entirely unaffected. The straddle is dangerous in **one** direction only. The whole discipline reduces to: **never let the client get ahead.**

`POST /v1/turso/token` retires on its own later cycle — unset `TURSO_PLATFORM_TOKEN` first so it answers its existing 503 and old clients keep their baked token, confirm zero traffic, then delete the handler.

---

## 8. Acceptance gates

| # | Gate | Check |
|---|---|---|
| P1 | No client read call sites | `grep -rn "\.remote_db\|\.log_db" --include=*.rs pollis-core/src pollis-tui/src src-tauri/src \| grep -v tests` → empty (baseline 102 + 12) |
| P2 | No token machinery | `grep -rn "spawn_turso_token_refresh\|set_remote_token\|ds_turso_token\|RemoteDb"` → empty outside DS + harness |
| P3 | **Credential-free build passes real flows** | build with all four unset; full `e2e-*` suite and `src-tauri/tests/flows` green against a real DS |
| P4 | Nothing leaked into the bytes | `strings <AppImage> \| grep -iE 'turso\|libsql://\|^eyJ'` → empty (the technique that found the baked `LOG_DB_TOKEN` in v1.8.4) |
| P5 | libsql gone from every client | `cargo tree -p pollis-tui -i libsql`, `-p pollis -i libsql`, android-target `pollis-core` → all "did not match any packages" |
| P6 | Advisory chain gone | `-i tonic`, `-i hyper@0.14`, `-i hyper-timeout`, `-i axum@0.6` → no match; `cargo deny check advisories` green with `deny.toml:48-50` removed |
| P7 | Over-deletion guard | `cargo tree -p pollis-delivery -i libsql` → **still matches** |
| P8 | Recipe consistency throughout | `check-build-recipe.py` exits 0 after **every** CI commit, not just at the end |
| P9 | Snapshot invariant | test that GroupInfo-present ⟹ Welcome-visible in one `conversation-state` response; and that a torn pair is not constructible |
| P10 | Contiguity invariant | test that a hole in a commit batch is impossible; gap-recovery still fires on a genuine retention prune via `pruned_below` |
| P11 | Regression tripwire | new test (sibling to `no_client_side_remote_writes.rs`) asserting `pollis-core` declares no `libsql` dependency — fails on a deliberate revert |
| P12 | Reproducibility gets stronger | fork with **zero** repo vars and **zero** secrets reproduces the first credential-free tag's Linux AppImage payload hash and matches the transparency log |

P12 is the payoff for `docs/reproducible-builds-residuals.md` §4, which currently names `LOG_DB_TOKEN` "the remaining secretless-repro blocker". Do not rewrite that doc before the run proves it — asserting it early is the exact failure the doc itself records.

---

## 9. Out of scope — file separately

- The `h2` / `hyper 0.14` advisory bump for the **DS** side (this ticket removes it from the *client* only)
- Voice-join `published_group_epoch` compares `epoch` without `generation`; after a suite migration the successor lineage's numerically-lower epoch reads as "not stale" and recovery silently never fires — same bug class already fixed in `group_info_is_stale`
- `user_groups` / `user_dms` are dead post-#540 and want a tightening migration to drop them
- `docs/goal-a-commit-log-sole-writer.md` §1b (R1–R6) is stale and should be corrected or marked historical
- `CLAUDE.md` line 22 — "Remote reads are direct to Turso (`state.remote_db`, SELECT only)" — must be rewritten **in this PR**, or the next agent re-adds a direct read

actuallydan:
Closes #987.

Every client-side remote read now goes through the Delivery Service, and the
client's database credential is deleted. Neither `pollis-core` nor the shipped
`pollis` binary links `libsql` any more, so a client-side query is a compile
error rather than a policy violation — which is what finally closes the residual
#917 recorded and could not fix (a read-only token is still a **whole-database**
token).

## Shape

102 `remote_db` call sites across 34 files and 12 `log_db` sites became ~22 new
`POST /v1/read/…` and `/v1/directory/…` endpoints. Wire types live in
`pollis-api` and are compiled by both ends.

| | |
|---|---|
| client reads | `pollis-core/src/commands/ds_reads.rs`, `.../mls/ds_reads.rs` |
| server reads | `pollis-delivery/src/{reads,directory,account_reads}.rs` |
| wire types | `pollis-api/src/{reads,directory,account_reads}.rs` |

Commits are one per phase (P1…P17); the PR is one ticket.

## The parts that are load-bearing

**One snapshot for the MLS control plane.** `POST /v1/read/conversation-state`
answers GroupInfo, the pending-Welcome flag, the lineage heads, the commit batch,
the retention floor and the two membership gates from **one read transaction**.
Split across requests, a device can see a GroupInfo without the Welcome
`submit_commit` wrote atomically beside it, external-join into its own inbound
Welcome, and be stranded permanently — and the Kani property
`ExternalJoin ⟹ !welcome_pending` stays true while becoming vacuous.

**Contiguity is reported, not enforced.** The DS checks the batch inside that
same transaction and reports a `hole` with both epochs. It deliberately does not
refuse: the client's response to a hole (drop the local group, external-join onto
the head) is the *recovery* — `epoch_gap_recovers_via_external_join` pins it —
and erroring would leave a genuinely damaged conversation permanently unreadable.
What the transaction removes is the **phantom** hole, where a batch and a head
read at two different instants manufactured a gap the log never had and sent a
client into that destructive recovery for nothing. `pruned_below` keeps a
legitimate retention prune distinguishable from damage.

**Reads are POST with signed bodies.** The canonical signing message covers the
path *without* its query string, so a signed `GET …?since=N&device_id=…` would
carry unauthenticated parameters — the #681 shape, where an unauthenticated
`GET /v1/commits` was a remote commit-log wipe. Read parameters live in the JSON
body, where `sha256(body)` covers them. GETs are also rate-limit-exempt and
absent from `pollis_api::ENDPOINTS`, so a GET read endpoint would be invisible to
the route-coverage test.

**Failure biases stay at their call sites.** `is_member` and `device_registered`
are `Option<bool>` on the wire and fail **CLOSED**; `welcome_pending` fails
**OPEN**. They are seven lines apart in `mls/group_state.rs`. HTTP makes those
failures common where a libsql read failure was rare, so collapsing them into one
"false on error" helper would silently invert one.

**Decisions stayed client-side.** `invariants::resolve` / `classify` /
`join_recovery` and `verify_added_devices` are unchanged. The DS supplies inputs
— cross-signing rows, commit bytes, membership facts — never verdicts.

**The bootstrap dead end (§6.5).** Three reads run with no device signature
because none exists at that moment:

- `POST /v1/auth/account-probe` — unauthenticated, tightest rate-limit tier. Runs
  pre-PIN at every launch: local DB closed (no signer), `device_id` unset, no OTP
  session. Answers `{ exists, has_identity, identity_version }` for a 128-bit
  ULID the caller read out of its own `accounts.json`. `identity_version` is on
  it because `ensure_device_cert` must stamp the version into the cert it is
  about to sign, and it runs *before* `mls_signature_pub` exists — the write that
  establishes the credential cannot be authenticated by that credential.
- `POST /v1/read/enrollment` — gated on possession of `request_id`. A
  session-gated poll is *guaranteed* to 401 first: both TTLs are 600s and the
  session is minted strictly earlier. The payload is sealed to an ephemeral
  X25519 key the id does not confer. Asking for the approval columns
  (`verification_code`) requires a full device signature.
- `POST /v1/read/recovery-blob` — OTP session only, security event per fetch.

## Credential removal

`RemoteDb`, `commands/turso_token.rs`, `state.remote_db` / `state.log_db`,
`ds_turso_token`, `query_only_view`, and the `TURSO_URL` / `TURSO_TOKEN` /
`LOG_DB_URL` / `LOG_DB_TOKEN` fields of `Config` are gone. So is
`POST /v1/turso/token` — with no client connection left, an endpoint that hands
out database credentials is pure attack surface — along with the
`TURSO_PLATFORM_TOKEN` / `TURSO_ORG` / `TURSO_DB` env it needed and nothing else
did.

All three libsql edges are cut. The direct `pollis-core` dependency; the
transitive one through `verifiable-log-serve → verifiable-log-builder`, now
behind `live` / `db-source` features that `pollis-core` does not enable; and
`src-tauri`'s own unconditional declaration, which existed only so the flows
harness could stand up a process-local "remote" and is now optional behind
`test-harness`. That last one is the edge that mattered for the shipped
artifact: `pollis` is the crate that becomes the installed binary, and it was
linking a database driver — plus the `hyper` 0.14 / `h2` 0.3 stack behind it —
to serve a test. The direct `hyper` 0.14 / `hyper-rustls` 0.25 / `tower-service`
declarations went too; they existed only to build the libsql-shaped SOCKS
connector.

What is left of libsql is server-side and correct: `pollis-delivery` and
`pollis-schema`'s migration runner *are* the database clients. That keeps one
advisory in the graph — RUSTSEC-2026-0258, `h2` 0.3 empty-DATA-frame DoS, no 0.3
backport — and it now has an `ignore` entry in `deny.toml` stating why: every
`h2` 0.3 instance is an *outbound* client to Turso over TLS, so the peer that
would have to send those frames is Turso; the DS's inbound listener, the surface
untrusted peers actually reach, is axum 0.8 on hyper 1.9, which is not built
with `h2` at all.

**The DS's own `TURSO_URL` / `TURSO_TOKEN` / `TURSO_ADMIN_TOKEN` /
`LOG_DB_ADMIN_TOKEN` are untouched** — the DS and the migration runner need them.
Only the client build stopped reading them.

CI edits respect `scripts/check-build-recipe.py`'s subset invariant
(`desktop-release.yml` ⊆ `rebuild-verify.yml` ⊆ `config.rs`): release workflows
first, then the rebuilder recipe, then `config.rs`. This also closes item 4 of
`docs/reproducible-builds-residuals.md` — the recipe is now six values, all of
them publishable, so a zero-secret third party can bit-reproduce the Linux
payload.

## New tests

- `pollis-delivery/tests/conversation_snapshot.rs` — P9/P10. A deterministic test
  shows a split read tears; a concurrent one asserts `snapshot` never does. Hole
  detection, prune/hole discrimination, and per-device Welcome scoping.
- `pollis-core/tests/no_client_side_remote_reads.rs` — P11. Walks
  `cargo tree -e normal` for both `pollis-core` (default features and
  `--all-features`) and `pollis` (the installed binary, default features), and
  fails if any database driver is in either graph by any path. Replaces the now
  ill-defined `query_only_view` runtime gate, which had nothing left to catch.
- `pollis-api::reads::check_contiguous` unit tests (pure).
- `pollis-delivery/tests/registered_devices.rs` — the #679 suite, moved to follow
  its query.
- `pollis-core/tests/resign_stale_certs.rs` and `tests/group_authz.rs` rewritten
  as pure tests of the predicates that stayed client-side; the SQL halves are
  covered server-side.

## Deployment ordering — DS FIRST

**The DS must be live in prod before any client release containing this change.**
A new client against an old DS has no fallback: every read 404s on a route that
does not exist yet, presenting as a client that signs in and appears to have no
groups, messages or devices. Deploy the DS (dev → verify → prod), confirm
`/version` shows the merged SHA, then tag the client release. The reverse
straddle is harmless. Recorded in `docs/deployments.md`.

## Also in this PR

`CLAUDE.md` line 22 rewritten. `.codesight/wiki/` (`commands.md`,
`secrets-broker.md`, `mls.md`, `database.md`, `testing.md`, `pollis-tui.md`),
`ARCHITECTURE.md`, `docs/security-whitepaper.md` §8, `docs/secrets-broker.md`,
`docs/deployments.md`, `docs/reproducible-builds-residuals.md`,
`docs/run-it-yourself.md`, `docs/pollis-tui-spec.md`,
`docs/relay-overlay-design.md`, `mobile/CLAUDE.md`, `README.md`, `.env.example`.

Push fan-out moved server-side (`pollis-delivery/src/push.rs`), which is what
took `push_token` — the most identifying row a client had any reason to read
about someone else — out of client visibility. Several post-write reads folded
into their write's response, removing both a round trip and a TOCTOU.

`derive_slug` moved to `pollis-api` so client and DS derive a slug from one
implementation.


actuallydan:
## Acceptance gates from #987 §8 — what was verified, and how

Verified against head `2794da32`. CI on that head is green on all 17 checks, `gate` included.

Legend: **MET** = checked, with the command below. **PARTIAL** / **NOT MET** / **DEPARTED** are spelled out — please read those four before merging.

---

### P1 — No client read call sites · **MET**

```
grep -rn "\.remote_db\|\.log_db" --include=*.rs pollis-core/src pollis-tui/src src-tauri/src | grep -v tests
```
→ empty (baseline was 102 + 12).

### P2 — No token machinery · **MET**

```
grep -rn "spawn_turso_token_refresh\|set_remote_token\|ds_turso_token\|RemoteDb" --include=*.rs .
```
→ no *symbol* survives. The remaining hits are prose in doc comments that name the deleted thing to explain why something looks the way it does (`net/overlay.rs`, `state.rs`, `pollis-relay/src/backoff.rs`, `pollis-delivery/src/db.rs`). P18–P20 rewrote the ones that still described it in the present tense.

### P3 — Credential-free build passes real flows · **PARTIAL**

```
env -u TURSO_URL -u TURSO_TOKEN -u LOG_DB_URL -u LOG_DB_TOKEN \
  cargo test -p pollis --no-default-features --features test-harness --test flows
→ test result: ok. 94 passed; 0 failed; 1 ignored; finished in 1684.40s
```
Nothing to unset in practice: `Config::for_test()` reads nothing from the environment any more, and P12 deleted the `.env.test` provisioning step from `mls-tests.yml`, so CI's `tests` job proves the same property on its own.

**What is missing:** the gate also asks for the `e2e-*` suite green **against a real DS**. That was not run — this sandbox has no deployed dev DS and no Turso credentials. The in-process DS the flows harness mounts is `pollis_delivery::build_router_with_state`, i.e. the production router, not a mock, so the dispatch path is real; but a real network hop to `api-dev` has not been exercised. **This should be done on dev before the client release is cut** — see "Deployment ordering" in the PR body.

### P4 — Nothing leaked into the bytes · **NOT MET (cannot run here)**

`strings <AppImage> | grep -iE 'turso|libsql://|^eyJ'` needs a built AppImage; the `media` feature does not build in this sandbox (below). The structural argument is that there is nothing left to bake — `Config` has no `TURSO_URL` / `TURSO_TOKEN` / `LOG_DB_URL` / `LOG_DB_TOKEN` field, so no `option_env!` reads them, and P8 confirms the release recipe is six values — but **please run the `strings` check on the first release artifact anyway.** It is the technique that caught the baked `LOG_DB_TOKEN` in v1.8.4, and a structural argument is not the same evidence.

### P5 — libsql gone from every client · **MET**

`cargo tree -e normal -i <crate>`, per package (so features resolve for that package, not the workspace union):

| package | libsql | libsql-hrana |
|---|---|---|
| `pollis-core` (default) | absent | absent |
| `pollis-core` (`--all-features`) | absent | absent |
| `pollis` — the installed binary | absent | absent |
| `pollis-tui` | absent | absent |

Mobile, `pollis-core --no-default-features` (media off, as mobile builds):

| target | libsql |
|---|---|
| `aarch64-linux-android` | absent |
| `armv7-linux-androideabi` | absent |
| `aarch64-apple-ios` | absent |

`pollis` needed a fix to reach this: `src-tauri/Cargo.toml` declared `libsql` **unconditionally** — used only by `src/test_harness.rs` and `tests/flows/` — so every shipped desktop build linked a database driver to serve a test. P17 made it `optional = true`, reached only through `test-harness`.

### P6 — Advisory chain gone from the client · **MET** · from the workspace · **DEPARTED**

Client half, all absent from `-e normal` for `pollis-core`, `pollis`, `pollis-tui`, and the three mobile targets: `tonic`, `hyper 0.14`, `hyper-timeout`, `axum 0.6`, `h2`.

**Where I departed from the ticket.** The gate also says `cargo deny check advisories` should pass with `deny.toml:48-50` (the libsql-attributed rustls-webpki ignores) **removed**. That is not achievable, and I think the gate as written is in direct tension with P7:

- `deny.toml` sets `[graph] all-features = true` and analyses the **whole workspace**;
- P7 requires `pollis-delivery` to keep libsql — it *is* the database client;
- so libsql, and the rustls-webpki 0.102 it pins, stay in the analysed graph. Those three ignores must stay.

A new advisory also landed during this work: **RUSTSEC-2026-0258**, `h2` 0.3 queues empty DATA frames without limit. Fixed in `h2` 0.4.16; no 0.3 backport, and `tonic` 0.11 pins `^0.3`. It reaches the graph through exactly one edge — `libsql → tonic 0.11 → hyper 0.14 → h2 0.3` — which is why cutting the third libsql edge (P17) was necessary but not sufficient. I added **one** documented `ignore` with the reachability argument: every `h2` 0.3 instance is an **outbound client to Turso over TLS**, so the peer that would have to send those frames is Turso or a broken TLS session; the DS's inbound listener, which is the surface untrusted peers actually reach, is axum 0.8 on hyper 1.9 and is **not built with h2 at all** (`cargo tree -p pollis-delivery -i h2` shows hyper 1.9 absent from the parents). Drop the entry when libsql moves to tonic 0.12+.

`supply-chain` was red on `main` before this PR and is **green** on this head.

### P7 — Over-deletion guard · **MET**

```
cargo tree -p pollis-delivery -i libsql
→ libsql v0.9.30
  ├── pollis-delivery
  └── pollis-schema → pollis-delivery
```
Still matches, as required.

### P8 — Recipe consistency throughout · **NOT MET at one commit**

At head, and at 20 of the 21 commits, `scripts/check-build-recipe.py` exits 0:

```
Build recipe OK — 6 value(s) baked by the release are all in the rebuilder's
recipe (6 entries), and every recipe entry is read by the client.
```
The other three checkers also pass: `check-ds-config-chain.py`, `check-pinned-log-key.py`, `check-workflows.py`.

**The miss.** The ticket mandated the edit order *release workflow → rebuilder recipe → `config.rs` last*, because the invariant is (release ⊆ rebuilder ⊆ config). I did it the wrong way round: `config.rs` lost the four fields in **P7 `4f5df4e6`**, and the workflows were not trimmed until **P8 `001379ae`**. I replayed the checker over every commit on the branch to find out rather than assume:

```
OK   d919582d P1  …  OK b99184b7 P6
FAIL 4f5df4e6 P7: delete the client's database credential
OK   001379ae P8  …  OK 2794da32 P21
```

At `4f5df4e6` the checker reports 4 problems: `rebuild-verify.yml` still passes `TURSO_URL` / `TURSO_TOKEN` / `LOG_DB_URL` / `LOG_DB_TOKEN`, which `config.rs` no longer reads.

Consequences, stated plainly: the merged tree is correct, and CI never saw it because every *pushed* head was at or past P8 — `shell-scripts` is green on all of them. But anyone bisecting the branch or cherry-picking P7 alone lands on a recipe that cannot reproduce. I have **not** rewritten history to fix it, because a force-push would invalidate the green run this PR is currently being judged on. Say the word and I will reorder the two commits and re-run CI.

### P9 — Snapshot invariant · **MET**

`cargo test -p pollis-delivery --test conversation_snapshot` → **7 passed; 0 failed**.

- `a_split_read_can_observe_a_torn_pair` — deterministic, proves the hazard is real by reading GroupInfo and the Welcome flag at two instants.
- `torn_group_info_and_welcome_is_not_observable` — `#[tokio::test(flavor = "multi_thread", worker_threads = 2)]`; a writer runs `submit_commit`'s atomic write in a loop while a reader samples `snapshot()`. Asserts `has_group_info == snap.welcome_pending` on every sample, **and** that both states were actually observed, so it cannot pass by never racing.
- `the_welcome_flag_is_scoped_to_the_asking_device` — the flag answers for the asking device, not the conversation.

### P10 — Contiguity invariant · **MET as re-derived** · **DEPARTED from the wording**

`a_contiguous_batch_is_served`, `a_hole_in_the_batch_is_reported`, `a_retention_prune_surfaces_as_pruned_below_not_as_an_error`, `an_empty_lineage_is_neither_a_hole_nor_a_prune`, plus pure unit tests on `pollis_api::reads::check_contiguous`.

The gate says "a hole in a commit batch is **impossible**". I implemented **report, not refuse**, and I want that on the record because it is a deliberate departure.

Refusing the batch breaks two existing tests that encode the product rule — `epoch_gap_recovers_via_external_join` and `commit_log_prune_recovers_via_external_join`. Both punch a real hole and expect the client to recover. The client's response to a hole (drop the local group, external-join onto the head) **is** the recovery; an error instead leaves a genuinely damaged conversation permanently unreadable, which "messages must work" forbids. What the single read transaction actually makes impossible is the **phantom** hole — a batch and a head read at two different instants manufacturing a gap the log never had and sending a client into that destructive recovery for nothing. `pruned_below` keeps a legitimate retention prune distinguishable from damage. The DB trigger `trg_mls_commit_log_no_forward_gap` still makes a *forward* gap unrepresentable on INSERT.

### P11 — Regression tripwire · **MET**

`pollis-core/tests/no_client_side_remote_reads.rs` → **3 passed**: `pollis_core_does_not_link_a_database_driver`, `no_feature_of_pollis_core_pulls_in_a_database_driver` (`--all-features`), `the_shipping_desktop_binary_does_not_link_a_database_driver` (added in P17, covers `pollis`). It walks `cargo tree` rather than scanning source, and carries sanity assertions so a `cargo tree` output change cannot make it pass by seeing nothing.

### P12 — Reproducibility gets stronger · **NOT MET (needs a release)**

Requires a fork with zero repo vars and zero secrets to reproduce the first credential-free tag's AppImage hash. Cannot be done pre-merge. What this PR does is remove the blocker: the recipe is six values, all publishable, which is `docs/reproducible-builds-residuals.md` item 4. That doc has been updated to say the blocker is removed — **not** that a secretless reproduction has been demonstrated. Per the doc's own warning, asserting it before the run is the exact failure it records.

---

## Environment limitation you should know about

**This sandbox cannot build the `media` feature at all.** `webrtc-audio-processing-sys` shells out to meson, which fetches abseil from `wrapdb.mesonbuild.com`; the agent proxy returns 403. So no local `cargo check`, `cargo test` or `cargo clippy` run in this session ever compiled `pollis-core/src/commands/voice_e2ee.rs` or `pollis-core/src/commands/livekit/identity.rs` — the two `media`-gated files this PR changes. **They were only ever compiled by CI.**

That gap produced a real defect. CI caught it: `voice_e2ee.rs:73` called `crate::commands::ds_reads::resolve_mls_group`, but that function lives in `crate::commands::mls::ds_reads`. Fixed in P21. After the fix I hand-checked every symbol and every caller in both files against their real definitions (`one_to_one_dm`, `users`, `conversation_snapshot`, `state_query`, `catch_up_mls_group_interleaved`, and the `&AppState` → `&Arc<AppState>` signature change against its four call sites in `voice/lifecycle.rs` and `livekit/participants.rs`), and CI has since compiled and run past that point. But the local verification loop had a blind spot over the voice path for most of this work, and reviewers should weight those two files accordingly.

## Also worth flagging

- **`published_group_epoch` compares `epoch` without `generation`** (`voice_e2ee.rs`). After a suite migration the successor lineage's numerically-lower epoch reads as "not stale", so voice-key recovery silently never fires. Pre-existing, out of scope here, documented in the function's doc comment — the same bug class `group_info_is_stale` already fixes for the republish path. Worth its own ticket.
- **Deployment order is load-bearing: DS to prod first.** A new client against an old DS has no fallback — every read 404s and presents as a client that signs in and appears to have no groups, messages or devices. Recorded in `docs/deployments.md` and in the PR body.


---

PR #993, opened 2026-08-20
Link SQLCipher on Windows with a RustCrypto crypto provider

### LINKED ISSUE #992: Windows local database is not encrypted at rest — SQLCipher has never been linked there
Split out of #991, which found it while fixing the v1.10.0 Windows link failure. #991 documents and pins the state; it deliberately does not attempt to close it.

## The state

On Windows, `pollis_{user_id}.db` is written in the clear. `PRAGMA key` is an *unknown* pragma to the sqlite3 that is actually linked, and SQLite ignores unknown pragmas silently — no error, no warning, every write succeeds. The `db_key` in the keystore exists and guards nothing. Decrypted message plaintext and all MLS group state (leaf private keys, signature keypairs, KeyPackage private halves) sit on disk unprotected.

This is not new. Until #988 the graph contained `libsql`, whose `libsql-ffi` links a second, plain sqlite3 amalgamation; it won the `sqlite3_*` symbols, so SQLCipher's object was never pulled into the Windows link. Every Windows build shipped to date has behaved this way. Linux, macOS and mobile are unaffected — verified by `sqlcipher_is_the_sqlite_we_actually_linked` in `pollis-core/src/db/local.rs`.

## Why it is not a config change

`bundled-sqlcipher` on Windows links a **static** libcrypto (there is no shared system OpenSSL). MSVC resolves a static archive by pulling whole object files, and SQLCipher built against OpenSSL 3 headers reaches the provider API (`EVP_CIPHER_fetch`), which BoringSSL does not implement. The first such symbol drags in OpenSSL's provider core and transitively 169 objects across X.509/PEM/RSA/EC — most of which the BoringSSL that `libwebrtc-sys` puts in the same image also defines. Result: 1,253 fatal `LNK2005` duplicate-symbol errors. ELF prefers one definition and moves on; MSVC does not. Vendoring OpenSSL makes it strictly worse.

Directions worth evaluating, none of them free:

- Link OpenSSL as a **DLL** (vcpkg `x64-windows` triplet) so archive members are per-symbol import thunks and nothing is duplicated. Ships `libcrypto-3-x64.dll` in the NSIS bundle, and leaves SQLCipher binding some symbols to BoringSSL and others to the DLL — needs real runtime validation, not just a green link.
- Give SQLCipher a non-OpenSSL crypto backend on Windows (CNG). SQLCipher upstream has CommonCrypto and LibTomCrypt backends but no CNG one, and `libsqlite3-sys` exposes neither.
- Encrypt at the application layer above rusqlite, platform-independently.

## Whatever the mechanism, a migration is mandatory

Every existing Windows database is plaintext. A build that starts encrypting will fail to open them, surface `SqliteFailure(NotADatabase)`, and trip the narrow wipe in `LocalDb::open_at` — destroying the device's message history *and* its MLS state, which forces a re-enrolment. That is a fourth unsanctioned history loss under the "messages must work" principle. Detect a plaintext file and `ATTACH ... KEY` + `sqlcipher_export` it before the first keyed open.

## Guard rail already in place

`sqlcipher_is_the_sqlite_we_actually_linked` asserts in both directions and runs on Linux (`mls-tests.yml`) and Windows (`windows-link.yml`). Its Windows arm asserts SQLCipher is **absent**, so the fix for this issue starts by flipping that assertion — which is the reminder that `docs/security-whitepaper.md` §7.0, the Cargo note and the migration all have to move together.

actuallydan:
Closes #992.

## What was linked, and what is linked now

Windows had `rusqlite`'s plain `bundled` feature: stock SQLite, no codec, `PRAGMA key` an unknown pragma that SQLite ignores silently, `pollis_{user_id}.db` written in the clear.

Windows now takes `rusqlite`'s **linked** `sqlcipher` feature (plus `modern_sqlite`, i.e. `libsqlite3-sys/bundled_bindings`, so the declarations come from the crate's pregenerated bindings and no header is needed). `libsqlite3-sys` therefore compiles no C at all on Windows and only emits `-l sqlcipher`; the new `pollis-sqlcipher` crate supplies that library.

Linux, macOS, iOS and Android are untouched — `cargo tree -e features -i libsqlite3-sys` resolves identically for all four.

## Why the crypto provider is RustCrypto and not OpenSSL

`bundled-sqlcipher*` on Windows links a static OpenSSL, MSVC resolves a static archive by pulling whole object files, and SQLCipher against OpenSSL 3 headers reaches `EVP_CIPHER_fetch` — which BoringSSL does not implement, so OpenSSL's provider core comes in and collides with the BoringSSL `libwebrtc-sys` already put in the image. 1,253 `LNK2005`.

Aiming SQLCipher at that BoringSSL instead was evaluated and rejected: BoringSSL changed `PKCS5_PBKDF2_HMAC` (and others) from `int` lengths to `size_t`. The calls would link and then pass 32-bit arguments into 64-bit parameters — a green build with undefined key derivation, which is the same failure mode as the bug being fixed. It also needs headers that only exist inside LiveKit's prebuilt WebRTC archive, and would tie disk encryption to the `media` feature.

So: no second libcrypto in the image at all. `pollis-sqlcipher` compiles the **stock, unpatched** SQLCipher amalgamation with `-DSQLCIPHER_CRYPTO_LIBTOMCRYPT` — the one upstream provider whose entire external surface is a single header we are free to supply — and `include/tomcrypt.h` declares the ~20 primitives it actually calls. `src/abi.rs` implements them over `aes`, `sha1`, `sha2`, `hmac`, `pbkdf2` and the OS CSPRNG. **All five are already in `Cargo.lock`; nothing new enters the supply chain.** No `/FORCE:MULTIPLE`, no linker flag picking a duplicate arbitrarily, no weakened cipher.

The output is unmodified SQLCipher 4 — AES-256-CBC pages, per-page HMAC-SHA512, PBKDF2-HMAC-SHA512, 256000 iterations — because a provider supplies primitives, not format. `PRAGMA cipher_provider` still answers `libtomcrypt` (upstream's string, in code we chose not to patch); `PRAGMA cipher_provider_version` answers `pollis-rustcrypto-1`.

`vendor/sqlite3.c` is copied byte-for-byte from `libsqlite3-sys` 0.35.0 with its SHA-256 recorded in `vendor/PROVENANCE.md`; the review surface is one hash and ~360 lines of provider, not nine megabytes. A stale Go-era `vendor/` rule in `.gitignore` needed a negation to check it in.

## Existing plaintext databases

**Recreated empty, not migrated**, and the plaintext is destroyed rather than unlinked. `LocalDb::open_at` now checks every existing file for the unencrypted `SQLite format 3\0` header and, on a hit, overwrites the file and its `-wal`/`-shm`/`-journal` sidecars with zeroes, fsyncs, and removes them.

`sqlcipher_export` was the alternative and was rejected: there is no production Windows user base to preserve anything for, the conversion has a window where both copies exist, and it only ends with no plaintext on disk if its cleanup also works. Starting empty is already a sanctioned loss ("a new device starts empty"), cannot half-complete, and leaves nothing behind. It is narrow (a short file, a non-database file and an already-encrypted file are all left alone) and idempotent. The overwrite is honestly best-effort on CoW/log-structured filesystems and wear-levelling SSDs; that is stated in the code.

Disposal is gated on this build actually linking SQLCipher, which is what makes it an upgrade rather than data loss — see the last section.

## Tests

`sqlcipher_is_the_sqlite_we_actually_linked` now requires SQLCipher on **every** platform. The `#[cfg(target_os = "windows")]` arm asserting its *absence* is gone and there is no `cfg` exception anywhere.

That check is necessary but not sufficient — it proves the library is present, not that the bytes are ciphertext. So `the_local_database_file_is_encrypted_at_rest` goes through `LocalDb::open_for_user` (the function the app calls, not a hand-built connection), writes a message with a distinctive marker, checkpoints, and then asserts against the raw file: the marker appears nowhere in it or its sidecars; the first 16 bytes are not `SQLite format 3\0`; an unkeyed connection cannot read the table; a wrong-keyed connection cannot either; the right key still can. `a_plaintext_database_is_destroyed_not_opened` covers the upgrade path and its idempotence.

`pollis-sqlcipher` additionally carries known-answer tests — RFC 6070 (PBKDF2-HMAC-SHA1), PBKDF2-HMAC-SHA512, RFC 4231/2202 (HMAC-SHA1/256/512), NIST SP 800-38A §F.2.5 (AES-256-CBC) — and `tests/sqlcipher.rs`, which drives real encrypted databases straight through the vendored library over hand-declared FFI (deliberately not through `rusqlite`, so no second sqlite3 is in that binary): the byte-level at-rest checks, wrong-key rejection, the SQLCipher 4 parameter set, and a 4000-row multi-page database round-tripped across a close so the per-page IVs and CBC chaining are exercised beyond page 1.

Its C is compiled on **every** target, not just Windows, purely so those tests run on the ordinary Linux gate: a mistake in the Windows-only crypto provider now fails in seconds instead of surviving until the 25-minute Windows job. Only Windows links it into the application.

CI: `mls-tests.yml` gains `cargo test -p pollis-sqlcipher`; `windows-link.yml` runs both that suite and the `db::local` tests natively on `windows-latest`, and greps for each load-bearing test name because libtest exits 0 when a filter matches nothing.

Verified on a real `windows-latest` MSVC runner: 9 + 5 + 16 tests pass and `pollis_core.dll` links with no duplicate symbols.

## The flows harness links a second sqlite3

Found by this PR's first Linux gate run, which failed three DM tests with `file is not a database`.

The integration harness runs the Delivery Service in-process, so it links `libsql` and therefore a second sqlite3 amalgamation. `nm` on the test binary confirms it: **no `sqlcipher_*` symbols at all** — `libsql-ffi`'s plain sqlite3 won the `sqlite3_*` symbols, SQLCipher's object was never pulled, and every client in that suite has always run on an unencrypted local database. Pre-existing, and exactly the ambiguity #992 is about, still alive in the test binary.

The new plaintext disposal turned that latent state fatal: it fired on every open, and in the multi-device tests — where two simulated devices of one user share a `pollis_{user_id}.db` — it zeroed the file under the other device's live connection.

Fixed by gating disposal on `sqlcipher_is_linked()`. Destroying a plaintext database is an upgrade when the replacement will be encrypted and pure data loss when it will not; in a codec-less binary it is an endless destroy-and-recreate. Production behaviour is unchanged, because a shipped build cannot be in that state — `sqlcipher_is_the_sqlite_we_actually_linked` now fails the build on every platform if it ever is. The probe runs on a connection the open path makes anyway rather than a scratch in-memory one, because `sqlite3_initialize()` at an arbitrary moment makes libsql's `sqlite3_config` panic (a hazard `.codesight/wiki/testing.md` already documents).

Rather than hide the harness's state, `tests/flows/linked_sqlite.rs` asserts it: a green flows run is not evidence about at-rest encryption in either direction. If `libsql` ever leaves that binary the test fails, and the right response is to delete it.

## Docs

Every statement that Windows is unencrypted is now false and has been rewritten: `docs/security-whitepaper.md` (trust table + §7.0, now a per-platform provider table, plus the honest note about what the flows harness does and does not cover), `ARCHITECTURE.md`, `docs/security-explained.md`, `website/security.html`, `.codesight/wiki/database.md`, `.codesight/wiki/testing.md`, plus `docs/deployments.md`, `CLAUDE.md`, `deny.toml` and the `desktop-release.yml` OpenSSL note.


actuallydan:
## Post-merge release checklist

Blast radius: `pollis-core/` (→ desktop, CLI, mobile), new `pollis-sqlcipher/` (Windows link only), `website/`. No wire contract, no schema, no relay, no transparency-log change.

- **Desktop app** — **defer.** Needs a `v*` tag, which is a release decision, not a merge one. Windows users get nothing from this until that release ships; the encryption is inert in `main`.
- **CLI (`pollis`)** — **defer**, same reason (`cli-release.yml` is `workflow_dispatch`). Rides `pollis-core`, so a Windows CLI build gets the same encrypted local DB whenever it is next cut.
- **DS** — **N/A.** No DS code and no mirrored wire contract: nothing in `pollis-api`, canonical signing, envelope encoding or the `/v1/…` surface moved. This is entirely client-side at-rest storage.
- **Relay pool** — **N/A.** `pollis-relay` / `pollis-device-cert` untouched.
- **Mobile** — **N/A**, and checked rather than assumed: `cargo tree -e features -i libsqlite3-sys` resolves identically for `aarch64-linux-android` and `aarch64-apple-ios` before and after (still `bundled-sqlcipher-vendored-openssl`), and `android-check`, `ios-check` and `android-build` are green on the merged commit. `pollis-sqlcipher` is a `cfg(windows)` dependency, so it is not in the mobile graph at all. No uniffi/`#[cfg]` follow-up.
- **pollis-verify CLI** — **N/A.** `verifiable-log*` untouched.
- **Transparency log** — **N/A.**
- **DB migration** — **N/A.** No migration added; nothing pending on prod.
- **Website** — **defer, deliberately.** `website/security.html` was rewritten, but `website-deploy.yml` is `workflow_dispatch` so nothing published automatically, and that is the right outcome: the live page describes what installed clients do, and the new text ("encrypted like everywhere else") does not become true for a Windows user until a desktop release carrying this ships. **Deploy the website with or after that release, not before it** — publishing it now would make pollis.com claim an at-rest property the shipped binary does not yet have.

Net: nothing is silently stale. Two real deploys are pending and both hang off the same event — the next desktop release — with the website deploy ordered behind it.


---

